PR #2, opened 2018-08-24
Update in Training function and added validate function - need to add more compatible metrics

priyatharsan:
I didnt do the previous changes yet. Thought I would do them together for both PRs altogether


examples/test.py
```diff
@@ -0,0 +1,58 @@
...
+from vulcanai2 import models, datasets
+
+import torch
+import torch.nn as nn
+import torchvision.transforms as transforms
+from torch.utils.data import DataLoader
+
+
+sys.path.append('../')
+normalize = transforms.Normalize(mean=[x/255.0 for x in [125.3, 123.0, 113.9]],
```
rfratila: why are these 3 values chosen for x?

examples/test.py
```diff
@@ -0,0 +1,58 @@
...
+                                     std=[x/255.0 for x in [63.0, 62.1, 66.7]])
+
+transform = transforms.Compose([transforms.ToTensor(),
+                                transforms.Normalize((0.1307,), (0.3081,))])
```
rfratila: why these normalization values?

priyatharsan:
this branch has been based off in this https://github.com/Aifred-Health/Vulcan2/pull/3

PR #5, opened 2018-09-10
dataset works as is

nirtiac:


nirtiac:
Merging dataset minor updates

vulcanai2/datasets/tabulardataset.py
```diff
@@ -67,41 +108,51 @@ def list_all_features(self):
...
+    def list_all_categorical_features(self):
         """
         Returns all columns that contain categorical values
         :return: all columns that contain categorical values
         """
-        return [key for key in dict(self.df.dtypes) if dict(self.df.dtypes)[key] not in ['float64', 'int64']]
+        return [key for key in dict(self.df.dtypes) if dict(self.df.dtypes)[key] not in ['float64', 'int64', 'float32', 'int32']]
 
     #TODO: check this doesn't operate in place... damn
     def remove_majority_null(self, threshold):
```
rfratila: With remove_majority_null and remove_unique, let's also make a note to the user of which columns were removed. This won't need to be the default but we can use our logger to specify the level of detail required from these functions. We want to be super transparent about what we remove and why when automating it. Looks great though :)

vulcanai2/datasets/tabulardataset.py
```diff
@@ -22,27 +25,27 @@ def __init__(self, data, labelColumn="label", joinColumn=None, indexList=None):
...
+        return self.df.shape[0]
 
     def __getitem__(self, idx):
         """
         Generates one sample of data
         """
-        #TODO: check types, possibly wrong
-        return list(self.df.drop(self.labelColumn, axis=1)[[idx]]), list(self.df[[self.labelColumn]].iloc[[idx]])[0]
+        #TODO: check types
+        return self.df.drop(self.labelColumn, axis=1).iloc[[2]].values.tolist()[0], self.df[[self.labelColumn]].iloc[[idx]].values.tolist()[0]
```
rfratila: What is this doing? seems pretty complicated to have on one line, maybe we should break it up?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -56,9 +59,47 @@ def save_dataframe(self, file_path):
...
+        """
+        Create one-hot encoding for all categorical features.
+        :param columnNames: All columns that you want to one-hot encode. You should probably use this if you have columns like patientID
+        :return:
+        """
+        self.df = pd.get_dummies(self.df, dummy_na=True, columns=columnNames)
+
+    #TODO: check cause this may cause problems with vars originally containing underscores
+    #taken from https://stackoverflow.com/questions/34523111/the-most-elegant-way-to-get-back-from-pandas-df-dummies
+    def reverse_create_dummies(self):
```
rfratila: doesn't this serve the same purpose as get_class()?

PR #11, opened 2018-09-14
added get_output_shape function

priyatharsan:
@rfratila


vulcanai2/tests/models/test_layers.py
```diff
@@ -3,25 +3,17 @@
 import torch
 from vulcanai2.models.layers import *
 
-class TestBaseLayer:
+class TestBaseUnit:
     @pytest.fixture
-    def baselayer(self): 
-        return BaseLayer()
-
-    # TODO: Needs fixing assigning values to NoneType
```
rfratila: Why was this deleted?

vulcanai2/models/utils.py
```diff
@@ -0,0 +1,62 @@
...
+def get_size(summary_dict, output):
+    if isinstance(output, tuple):
+        for i in xrange(len(output)):
+            summary_dict[i] = OrderedDict()
+            summary_dict[i] = get_size(summary_dict[i],output[i])
+    else:
+        summary_dict['output_shape'] = list(output.size())
+    return summary_dict
+ 
+def get_output_shapes(input_size, model):
```
rfratila: I would like to see more documentation here since it seems quite complicated to extract the output shapes of each layer. Also if a function is in a class, we shouldn't need to pass it a model since it should just look at `self`. it doesn't make sense to use it as a static class when we've already instantiated it in the script that it is made it.

vulcanai2/tests/models/test_layers.py
priyatharsan: Rename to `BaseUnit`

rfratila: sorry, I was referring to the #TODO comment below

rfratila: the test function of modifying variables in baseunit

vulcanai2/models/utils.py
priyatharsan: Tbh it was quite complicate for me to understand as well. 
ref links: https://pytorch.org/tutorials/beginner/former_torchies/nn_tutorial.html#forward-and-backward-function-hooks
https://github.com/sksq96/pytorch-summary/blob/master/torchsummary/torchsummary.py

vulcanai2/tests/models/test_layers.py
priyatharsan: those variables are dummy and are not being used. Although I would like to take a look at them once while working on the other layer classes.

vulcanai2/models/utils.py
rfratila: ok for now let's put a link to this resource in the code and modify the function signature to just use the internal model instead of having to pass it one

vulcanai2/tests/models/test_layers.py
rfratila: ok let's make a not of that through docs in the BaseUnit class

vulcanai2/models/utils.py
priyatharsan: so in that case it has to be moved to the `BaseNetwork`, right?

rfratila: yeah good idea. it will be used in all networks

vulcanai2/models/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .basenetwork import BaseNetwork
 from .cnn import ConvNet
 from .dnn import DenseNet
+from .utils import *
```
nirtiac: Why do we have this? I'm worried about "polluting the namespace", especially if our utility functions are generically named.

rfratila: that is a good point. @priyatharsan would you be able to specify what from utils you require so we can be more specific with the imports?

rfratila: also on that note, I think what is found in utils should be tied to BaseNetwork since intuitively you should be able to do model.get_size and such instead of passing a model to get_size(model) and get_output_shapes(model)? I'm curious to know your thoughts on this matter

nirtiac: Oh I didn't notice that. Good point, I agree completely, those should be with BaseNetwork.

priyatharsan: removed the contents of utils to BaseNetwork as it makes more sense there

rfratila: feel free to push your commit to this branch so we can look at the updated version

priyatharsan:
I'm merging the branch for now. Getting few conflicts locally.

rfratila:
can we delete the branch?

PR #17, opened 2018-09-16
added plotting records functionality

priyatharsan:
More plotting functionality needed to be added.

### DIFF
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -337,15 +337,12 @@ class BaseNetwork(nn.Module):
         try:
             if plot is True:
                 fig_number = plt.gcf().number + 1 if plt.fignum_exists(1) else 1
-
+                plt.show()
+                
             for epoch in trange(self.epoch, epochs, desc='Epoch: ', ncols=80):
-                epoch_time = time.time()
 
                 train_loss, train_acc = self._train_epoch()
-                train_epoch_time = time.time() - epoch_time
-
                 valid_loss, valid_acc = self._validate()
-                valid_epoch_time = time.time() - epoch_time
 
                 tqdm.write("\n Epoch {}:\n"
                            "Train Loss: {:.6f} | Test Loss: {:.6f} |"
@@ -367,6 +364,7 @@ class BaseNetwork(nn.Module):
                     plt.ion()
                     plt.figure(fig_number)
                     display_record(record=self.record)
+
         except KeyboardInterrupt:
             print("\n\n**********KeyboardInterrupt: Training stopped prematurely.**********\n\n")
 
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -19,9 +19,12 @@ import os
 import pickle
 import time
 from collections import OrderedDict
+
 import matplotlib
 import matplotlib.pyplot as plt
+import seaborn as sns
 
+sns.set()
 logger = logging.getLogger(__name__)
 
 
@@ -137,18 +140,6 @@ class BaseNetwork(nn.Module):
     @stopping_rule.setter
     def stopping_rule(self, value):
         self._stopping_rule = value
-
-    @property
-    def criter_spec(self):
-        """
-        Returns the criterion spec.
-        :return: The criterion spec.
-        """
-        return self._stopping_rule
-
-    @criter_spec.setter
-    def criter_spec(self, value):
-        self._criter_spec = value
 
     def get_flattened_size(self, network):
         """
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -1,17 +1,26 @@
 # -*- coding: utf-8 -*-
 """Defines the basenetwork class"""
+# Core imports
 import abc
 from torch.autograd import Variable
 import torch.nn.modules.loss as loss
+
+# Vulcan imports
 from .layers import *
 from .metrics import Metrics
+from vulcanai2.plotters.visualization import display_record
+
+# Generic imports
 import pydash as pdash
 from tqdm import tqdm, trange
 from datetime import datetime
 import logging
 import os
 import pickle
+import time
 from collections import OrderedDict
+import matplotlib
+import matplotlib.pyplot as plt
 
 logger = logging.getLogger(__name__)
 
@@ -72,6 +81,7 @@ class BaseNetwork(nn.Module):
         self.epochs = None
         self.retain_graph = False
         self.valid_interv = None
+        self.record = None
 
         self._create_network()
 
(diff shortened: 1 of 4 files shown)


vulcanai2/models/basenetwork.py
```diff
@@ -1,17 +1,26 @@
...
 """Defines the basenetwork class"""
+# Core imports
 import abc
 from torch.autograd import Variable
 import torch.nn.modules.loss as loss
+
+# Vulcan imports
 from .layers import *
 from .metrics import Metrics
+from vulcanai2.plotters.visualization import display_record
```
rfratila: you should do relative imports here

vulcanai2/models/basenetwork.py
```diff
@@ -261,6 +271,14 @@ def get_weights(self):
         """
         return self.state_dict()
 
+    def print_model_structure(self, input_dim=((1, 28, 28))):
```
rfratila: we shouldn't default the input_dim

vulcanai2/models/basenetwork.py
```diff
@@ -307,18 +326,49 @@ def fit(self, train_loader, val_loader, epochs, retain_graph=None, valid_interv=
...
-                           "Train Loss: {:.6f} | Test Loss: {:.6f} | "
-                           "Train Acc: {:.4f} | Test Acc: {:.4f}".format(
+                           "Train Loss: {:.6f} | Train Acc: {:.4f} in {:.2f}s \n"
+                           "Test Loss: {:.6f} | Test Acc: {:.4f} in {:.2f}s".format(
                                 epoch,
                                 train_loss,
-                                valid_loss,
                                 train_acc,
-                                valid_acc))
+                                train_epoch_time,
```
rfratila: Why are you also putting in the time if tqdm will tell you how much time is left in that epoch?

vulcanai2/plotters/visualization.py
```diff
@@ -21,350 +20,74 @@
...
-        matplotlib.use('Agg')
-else:
-    matplotlib.use('TkAgg')
 import matplotlib.pyplot as plt
 
 import logging
 logger = logging.getLogger(__name__)
 
 
-#TODO: update to work with pytorch
```
rfratila: Why did you delete all this stuff? it will need to be implemented. did you move it elsewhere?

vulcanai2/models/basenetwork.py
rfratila: I would prefer to have the epoch results on one line

vulcanai2/models/basenetwork.py
priyatharsan: something like `from ..plotters.visualization import display_record` ?

vulcanai2/models/basenetwork.py
priyatharsan: 👍

vulcanai2/models/basenetwork.py
rfratila: yeah

vulcanai2/models/basenetwork.py
priyatharsan: Makes sense. It was there in Vulcan so I thought I would add it. I will remove

rfratila: Fair enough. yeah Vulcan's command line info wasn't as robust as tqdm which is why I had it before but this new version works better

vulcanai2/plotters/visualization.py
priyatharsan: We just need to add plotting stuff here. I was working on saliency map got stuck on that the whole day. since I have little knowledge on that. Some functions (`round_list`, `get_confusion_matrix`) are already moved to metrics, some like `get_timestamp` are not necessary in visualization, unless if you'd like them to have?

rfratila: the most important ones from there are all the functions starting with display (e.g. display_tsne, _plot_reduction, display_pca, display_saliency_map) and stuff like that

rfratila:
When training ends, the graphs disappear. Try to "freeze" them in a way so they don't close once the training terminates

priyatharsan:
yeah right! I will work on that.

vulcanai2/models/basenetwork.py
```diff
@@ -261,6 +271,14 @@ def get_weights(self):
         """
         return self.state_dict()
 
+    def print_model_structure(self, input_dim):
```
rfratila: actually, since this is a part of basenetwork, do we store the input dimensions anywhere in this class? perhaps we can default to that.

vulcanai2/models/basenetwork.py
```diff
@@ -307,18 +326,47 @@ def fit(self, train_loader, val_loader, epochs, retain_graph=None, valid_interv=
...
+            validation_error=[],
+            validation_accuracy=[]
+        )
+
         try:
+            if plot is True:
+                fig_number = plt.gcf().number + 1 if plt.fignum_exists(1) else 1
+
             for epoch in trange(self.epoch, epochs, desc='Epoch: ', ncols=80):
+                epoch_time = time.time()
```
rfratila: don't need to calculate these anymore since tqdm does it for us

vulcanai2/models/basenetwork.py
```diff
@@ -307,18 +326,47 @@ def fit(self, train_loader, val_loader, epochs, retain_graph=None, valid_interv=
...
+
                 train_loss, train_acc = self._train_epoch()
+                train_epoch_time = time.time() - epoch_time
```
rfratila: don't need this either

vulcanai2/models/basenetwork.py
```diff
@@ -307,18 +326,47 @@ def fit(self, train_loader, val_loader, epochs, retain_graph=None, valid_interv=
...
+
                 valid_loss, valid_acc = self._validate()
+                valid_epoch_time = time.time() - epoch_time
```
rfratila: nor this

vulcanai2/plotters/visualization.py
```diff
@@ -21,18 +20,75 @@
...
-        matplotlib.use('Agg')
-else:
-    matplotlib.use('TkAgg')
 import matplotlib.pyplot as plt
 
 import logging
 logger = logging.getLogger(__name__)
 
 
+def display_record(record=None, load_path=None):
```
rfratila: lets try to make these a tiny bit prettier. perhaps we can use seaborn?

vulcanai2/models/basenetwork.py
priyatharsan: yeah good idea!

vulcanai2/plotters/visualization.py
```diff
@@ -248,4 +246,31 @@ def display_record(record=None, load_path=None, interactive=True):
 #     sal_map = sal_fun(input_data)
 #     if sal_map.shape != input_data.shape:
 #         raise ValueError('Shape mismatch')
-#     return sal_map
\ No newline at end of file
+#     return sal_map
+
+def Saliency_map(image,model,preprocess,ground_truth,use_gpu=False,method=util.GradType.GUIDED):
```
rfratila: Saliency map should not be specific to just images. I realize that we only tested it for image based datasets but it will need to be generalized to any incoming datatype. Let's find a better way to send data to this folder. lets have this function signature start with (network, input_x, input_y,...). Im not sure what datatype the variable for `preprocess` is but I think it may stray away from the scope of this function. simply, we just have the input, put it through and get the back gradients, and return those. Let me know if you have any questions

vulcanai2/plotters/visualization.py
priyatharsan: resolved! (partially I guess)

priyatharsan:
merging this PR, feel free to pitch in comments/reviews in this PR. The changes will be made in a new PR.

PR #20, opened 2018-10-03
Early stopping name fix

rfratila:
Fixed the early stopping name from the last PR. also patch the metric reset error.


priyatharsan:
lgtm 👍

PR #22, opened 2018-10-03
Multiple input functionality within BaseNetwork

priyatharsan:
Discussion needed on this. 

@rfratila I think this is how you were expecting our forward function. But I still wonder how are we running `train` and `validate` for multiple inputs? Not sure the forward function is running yet. I will have to run test on single input first. This PR is just for discussion.

### DIFF
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -40,7 +40,7 @@ class BaseNetwork(nn.Module):
     # TODO: reorganize these.
     def __init__(self, name, dimensions, config, save_path=None, input_networks=None, num_classes=None, 
                  activation=nn.ReLU(), pred_activation=nn.Softmax(dim=1), optim_spec={'name': 'Adam', 'lr': 0.001},
-                 lr_scheduler=None, early_stopping=None, criter_spec=nn.CrossEntropyLoss):
+                 lr_scheduler=None, early_stopping=None, criter_spec=nn.CrossEntropyLoss()):
         """
         Defines the network object.
         :param name: The name of the network. Used when saving the file.
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -18,7 +18,7 @@ import logging
 import os
 import pickle
 import time
-from collections import OrderedDict
+from collections import OrderedDict as odict
 import numpy as np
 
 import matplotlib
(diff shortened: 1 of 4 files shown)


vulcanai2/models/basenetwork.py
```diff
@@ -281,6 +285,19 @@ def _create_network(self):
         :return: None
         """
         pass
+    
+    def forward(self, *inputs):
```
nirtiac: I think I didn't know what was going on in this discussion. why is forward in base_network???

vulcanai2/models/basenetwork.py
```diff
@@ -281,6 +285,19 @@ def _create_network(self):
...
+        """Perform a forward pass through the module/modules."""
+
+        if isinstance(self.input_network, list):
+            outputs = []
+            for module in self.input_network:
+                outputs.append(module._forward(inputs))
+            network_output = torch.cat(outputs, 1)
+        else: 
+            network_output = self._forward(inputs)
+
```
nirtiac: here we would feed both through something like torch.cat((x,y),1)?

vulcanai2/models/basenetwork.py
```diff
@@ -281,6 +285,19 @@ def _create_network(self):
...
         """
         pass
+    
+    def forward(self, *inputs):
+        """Perform a forward pass through the module/modules."""
+
+        if isinstance(self.input_network, list):
+            outputs = []
+            for module in self.input_network:
+                outputs.append(module._forward(inputs))
```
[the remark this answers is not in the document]
priyatharsan: You are right! Like I mentioned this is not a working program yet I have to fix the code but before that I wanted to have a discussion if this is how @rfratila wanted.

vulcanai2/models/basenetwork.py
priyatharsan: the discussion is in our slack group

vulcanai2/models/basenetwork.py
priyatharsan: not for both we don't need to concatenate when we have only one input.

vulcanai2/models/basenetwork.py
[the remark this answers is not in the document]
rfratila: yeah we're on the right track. While we should have some sort of "get forward through all input networks" type of function, we can have it as a sister function to `forward` (which is specific to each module) in base network instead of overriding forward from basenetwork. Also I've been thinking more about the self.input_networks object and it should be a dict so we can attach options like freezing the weights of specific inputs @priyatharsan

vulcanai2/models/cnn.py
```diff
@@ -74,9 +74,9 @@ def _create_network(self):
     def _create_classification_layer(self, dim):
         self.network_tails = nn.ModuleList([DenseUnit(dim, out_d) for out_d in self.out_dim])
 
-    def forward(self, x):
+    def _forward(self, x):
```
rfratila: don't we need to have a `forward` function instead of the `_forward`? probably doesn't make much of a difference.

vulcanai2/models/basenetwork.py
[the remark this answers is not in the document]
rfratila: regarding @nirtiac 's comment, that would be one way to do it. The most important thing to keep in mind is making sure the right data inputs goes into the correct module.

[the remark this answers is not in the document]
rfratila: we will have to be very careful with the `outputs` list since the outputs from networks may not immediately compatible for concatenation. say we have a Conv output of shape `[16, 64, 28, 28]` and a dense output of shape `[16, 400]`, in this scenario we should probably linearize the conv output if it is feeding into a dense net (like the way you currently have it). Something else to think about is if you wanted a Conv + dense to feed into conv, do you maintain the conv output of the input network and just append the dense outputs to the shape of conv? does that make sense? im curious what you are thinking @nirtiac @priyatharsan

priyatharsan:
Moved to https://github.com/Aifred-Health/Vulcan2/pull/29

PR #23, opened 2018-10-05
Save function

nirtiac:
Will rebase momentarily!

Contains a basic working version of the save function; will use _state_dict in the future to allow for device switching, but adding this for now so that Priya can integrate the changes to what are instance variables.

### DIFF
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -76,6 +76,8 @@ class BaseNetwork(nn.Module):
         
         self.optim = None
         self.criterion = None
+        self.epoch = 0
+
         #self._itr = 0 #TODO: ?
 
         self._create_network()
@@ -302,8 +304,6 @@ class BaseNetwork(nn.Module):
 
         self._init_trainer(train_loader)
 
-        epoch = 0
-
         record = dict(
             epoch=[],
             train_error=[],
@@ -317,7 +317,7 @@ class BaseNetwork(nn.Module):
                 fig_number = plt.gcf().number + 1 if plt.fignum_exists(1) else 1
                 plt.show()
                 
-            for epoch in trange(epoch, epochs, desc='Epoch: ', ncols=80):
+            for epoch in trange(self.epoch, epochs, desc='Epoch: ', ncols=80):
 
                 train_loss, train_acc = self._train_epoch(train_loader, retain_graph)
                 valid_loss, valid_acc = self._validate(val_loader, retain_graph)
@@ -325,14 +325,14 @@ class BaseNetwork(nn.Module):
                 tqdm.write("\n Epoch {}:\n"
                            "Train Loss: {:.6f} | Test Loss: {:.6f} |"
                            "Train Acc: {:.4f} | Test Acc: {:.4f}".format(
-                                epoch,
+                                self.epoch,
                                 train_loss,
                                 valid_loss,
                                 train_acc,
                                 valid_acc
                                 ))
 
-                record['epoch'].append(epoch)
+                record['epoch'].append(self.epoch)
                 record['train_error'].append(train_loss)
                 record['train_accuracy'].append(train_acc)
                 record['validation_error'].append(valid_loss)
@@ -342,6 +342,8 @@ class BaseNetwork(nn.Module):
                     plt.ion()
                     plt.figure(fig_number)
                     #display_record(record=record)
+
+                self.epoch += 1
 
         except KeyboardInterrupt:
             print("\n\n**********KeyboardInterrupt: Training stopped prematurely.**********\n\n")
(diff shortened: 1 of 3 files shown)


vulcanai2/models/basenetwork.py
```diff
@@ -289,12 +285,10 @@ def _get_criterion(criterion_spec):
...
         return criterion_class(**criterion_spec)
 
-    def _init_trainer(self):
+    def _init_trainer(self, train_loader):
         self.optim = self._init_optimizer(self._optim_spec)
         self.criterion = self._get_criterion(self._criter_spec)
-
-        self.valid_interv = 2*len(self.train_loader)
-        self.epoch = 0
+        self.valid_interv = 2*len(train_loader)
```
rfratila: we can remove the self.valid_interv since it is now just passed to .fit. would simplify the function signature to remove it as well

vulcanai2/models/basenetwork.py
```diff
@@ -304,20 +298,13 @@ def fit(self, train_loader, val_loader, epochs,
         :param val_loader: The DataLoader object containing the validation data
         :param epochs: The number of epochs
         :param retain_graph: Specifies whether retain_graph will be true when .backwards is called.
-        :param valid_interv:
+        :param valid_interv: Specifies when validation should occur. Not yet implemented.
```
rfratila: I believe @priyatharsan implemented it recently. We just need to fix how we omit printing validation scores that don't exist in-between the interval

vulcanai2/models/basenetwork.py
```diff
@@ -304,20 +298,13 @@ def fit(self, train_loader, val_loader, epochs,
...
 
-        self.train_loader = train_loader
-        self.val_loader = val_loader
-        self.epochs = epochs
-        self.retain_graph = retain_graph
-        if valid_interv:
-            self.valid_interv = valid_interv
-
-        self._init_trainer()
+        self._init_trainer(train_loader)
```
rfratila: once we remove the self.valid_interv we can remove passing the train loader

vulcanai2/models/basenetwork.py
```diff
@@ -304,20 +298,13 @@ def fit(self, train_loader, val_loader, epochs,
...
 
-        self.record = dict(
+        record = dict(
```
rfratila: don't we want to hold onto the record? if we load up a model and continue to train it we should be able to see it's previous training cycles through the record @nirtiac . What are your thoughts on this?

vulcanai2/models/basenetwork.py
```diff
@@ -332,41 +319,43 @@ def fit(self, train_loader, val_loader, epochs,
...
+                record['train_error'].append(train_loss)
+                record['train_accuracy'].append(train_acc)
+                record['validation_error'].append(valid_loss)
+                record['validation_accuracy'].append(valid_acc)
 
                 if plot is True:
                     plt.ion()
                     plt.figure(fig_number)
-                    display_record(record=self.record)
+                    #display_record(record=record)
```
rfratila: why was this commented out?

vulcanai2/models/basenetwork.py
```diff
@@ -465,46 +454,75 @@ def forward_pass(self, input_data, convert_to_class=False):
...
-                                                          self.name))
-        save_name = '{}.network'.format(file_path)
-        print('Saving model as: {}'.format(save_name))
+        print("in save", self.epoch)
+
+        if not save_path:
+            save_path = "{date:%Y-%m-%d_%H:%M:%S}/".format(name=self.name, date=datetime.now())
+            logger.info("No save path provideded, saving to {}".format(save_path))
+
+        if not save_path.endswith("/"): #TODO: does this break windows?? no idea.
```
rfratila: @priyatharsan see if it does?

vulcanai2/models/basenetwork.py
```diff
@@ -465,46 +454,75 @@ def forward_pass(self, input_data, convert_to_class=False):
...
+        model_file_path = module_save_path + "model.pkl"
+        state_dict_file_path = module_save_path + "state_dict.pkl"
+        pickle.dump(self, open(model_file_path, "wb"), 2)
+        pickle.dump(self.state_dict, open(state_dict_file_path, "wb"), 2) #TODO: pretty sure this isn't necessary
 
+    #TODO: update the state dict and push to the appropriate device.
+    #https://pytorch.org/tutorials/beginner/saving_loading_models.html#what-is-a-state-dict
     @classmethod
-    def load_model(cls, load_path):
+    def load_ensemble(cls, load_path):
```
rfratila: let's generalize it to be just load_model() so it should be able to load individual and multi-component networks

priyatharsan:
@rfratila this PR is moved to https://github.com/Aifred-Health/Vulcan2/pull/24

rfratila:
oops sorry. I'll move over my comments

PR #24, opened 2018-10-05
new save function and edits to what is a class variable

nirtiac:
Contains a basic working version of the save function; will use _state_dict in the future to allow for device switching, but adding this for now so that Priya can integrate the changes to what are instance variables.

### DIFF
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -468,8 +468,6 @@ class BaseNetwork(nn.Module):
         :return: save path, for recursive purposes
         """
 
-        print("in save", self.epoch)
-
         if not save_path:
             save_path = "{date:%Y-%m-%d_%H:%M:%S}/".format(name=self.name, date=datetime.now())
             logger.info("No save path provideded, saving to {}".format(save_path))
@@ -510,7 +508,6 @@ class BaseNetwork(nn.Module):
 
         instance = pickle.load(open(model_file_path, 'rb'))
 
-        print("linstance epoch", instance.epoch)
         return instance
 
         # my_tensor = my_tensor.to(torch.device('cuda')).
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -8,7 +8,7 @@ import torch.nn.modules.loss as loss
 # Vulcan imports
 from .layers import *
 from .metrics import Metrics
-# from ..plotters.visualization import display_record
+from ..plotters.visualization import display_record
 
 # Generic imports
 import pydash as pdash
@@ -83,6 +83,14 @@ class BaseNetwork(nn.Module):
         self.optim = None
         self.criterion = None
         self.epoch = 0
+
+        self.record = dict(
+            epoch=[],
+            train_error=[],
+            train_accuracy=[],
+            validation_error=[],
+            validation_accuracy=[]
+        )
 
         # self._itr = 0 #TODO: ?
 
@@ -290,10 +298,9 @@ class BaseNetwork(nn.Module):
     def _init_criterion(criterion_spec):
         return criterion_spec()
 
-    def _init_trainer(self, train_loader):
+    def _init_trainer(self):
         self.optim = self._init_optimizer(self._optim_spec)
         self.criterion = self._init_criterion(self._criter_spec)
-        self.valid_interv = 2 * len(train_loader)
 
     def fit(self, train_loader, val_loader, epochs,
             retain_graph=None, valid_interv=4, plot=False):
@@ -303,19 +310,11 @@ class BaseNetwork(nn.Module):
         :param val_loader: The DataLoader object containing the validation data
         :param epochs: The number of epochs
         :param retain_graph: Specifies whether retain_graph will be true when .backwards is called.
-        :param valid_interv: Specifies when validation should occur. Not yet implemented.
+        :param valid_interv: Specifies the number of epochs before validation occurs.
         :return: None
         """
 
-        self._init_trainer(train_loader)
-
-        record = dict(
-            epoch=[],
-            train_error=[],
-            train_accuracy=[],
-            validation_error=[],
-            validation_accuracy=[]
-        )
+        self._init_trainer()
 
         try:
             if plot is True:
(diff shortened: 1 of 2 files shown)


vulcanai2/models/basenetwork.py
```diff
@@ -291,31 +290,26 @@ def _init_optimizer(self, optim_spec):
     def _init_criterion(criterion_spec):
         return criterion_spec()
 
-    def _init_trainer(self):
+    def _init_trainer(self, train_loader):
         self.optim = self._init_optimizer(self._optim_spec)
         self.criterion = self._init_criterion(self._criter_spec)
-        self.epoch = 0
+        self.valid_interv = 2 * len(train_loader)
```
rfratila: we can remove the self.valid_interv since it is now just passed to .fit. would simplify the function signature to remove it as well

vulcanai2/models/basenetwork.py
```diff
@@ -291,31 +290,26 @@ def _init_optimizer(self, optim_spec):
...
+    def fit(self, train_loader, val_loader, epochs,
             retain_graph=None, valid_interv=4, plot=False):
         """
         Trains the network on the provided data.
         :param train_loader: The DataLoader object containing the training data
         :param val_loader: The DataLoader object containing the validation data
         :param epochs: The number of epochs
         :param retain_graph: Specifies whether retain_graph will be true when .backwards is called.
-        :param valid_interv:
+        :param valid_interv: Specifies when validation should occur. Not yet implemented.
```
rfratila: I believe @priyatharsan implemented it recently. We just need to fix how we omit printing validation scores that don't exist in-between the interval

vulcanai2/models/basenetwork.py
```diff
@@ -291,31 +290,26 @@ def _init_optimizer(self, optim_spec):
...
         :return: None
         """
 
-        self.train_loader = train_loader
-        self.val_loader = val_loader
-        self.epochs = epochs
-        self.retain_graph = retain_graph
-
-        self._init_trainer()
+        self._init_trainer(train_loader)
```
rfratila: once we remove the self.valid_interv we can remove passing the train_loader

vulcanai2/models/basenetwork.py
```diff
@@ -291,31 +290,26 @@ def _init_optimizer(self, optim_spec):
...
 
-        self.record = dict(
+        record = dict(
```
rfratila: don't we want to hold onto the record as a class variable @nirtiac ? if we load up a model and continue to train it we should be able to see it's previous training cycles through the record. What are your thoughts on this?

vulcanai2/models/basenetwork.py
```diff
@@ -327,47 +321,49 @@ def fit(self, train_loader, val_loader, epochs,
...
+                record['train_error'].append(train_loss)
+                record['train_accuracy'].append(train_acc)
+                record['validation_error'].append(valid_loss)
+                record['validation_accuracy'].append(valid_acc)
 
                 if plot is True:
                     plt.ion()
                     plt.figure(fig_number)
-                    display_record(record=self.record)
+                    # display_record(record=record)
```
rfratila: Why was this commented out?

vulcanai2/models/basenetwork.py
```diff
@@ -465,46 +461,66 @@ def forward_pass(self, input_data, convert_to_class=False):
...
-                                                          self.name))
-        save_name = '{}.network'.format(file_path)
-        print('Saving model as: {}'.format(save_name))
+        if not save_path:
+            save_path = "{date:%Y-%m-%d_%H:%M:%S}/".format(name=self.name, date=datetime.now())
+            logger.info("No save path provideded, saving to {}".format(save_path))
 
-        with open(save_name, 'wb') as f:
-            pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)
+        if not save_path.endswith("/"):  # TODO: does this break windows?? no idea.
```
rfratila: @priyatharsan see if it does?

vulcanai2/models/basenetwork.py
```diff
@@ -465,46 +461,66 @@ def forward_pass(self, input_data, convert_to_class=False):
...
+        model_file_path = module_save_path + "model.pkl"
+        state_dict_file_path = module_save_path + "state_dict.pkl"
+        pickle.dump(self, open(model_file_path, "wb"), 2)
+        pickle.dump(self.state_dict, open(state_dict_file_path, "wb"), 2)  # TODO: pretty sure this isn't necessary
+
+    # TODO: update the state dict and push to the appropriate device.
+    # https://pytorch.org/tutorials/beginner/saving_loading_models.html#what-is-a-state-dict
     @classmethod
-    def load_model(cls, load_path):
+    def load_ensemble(cls, load_path):
```
rfratila: let's generalize this load function to be just load_model() so it should be able to load individual and multi-component networks

rfratila: maybe we can revisit torch.save to see if that might be useful?

vulcanai2/models/basenetwork.py
nirtiac: yeah, no that does make sense, didn't think that through. I'll add it back.

vulcanai2/models/basenetwork.py
nirtiac: oops. I couldn't get it to work on my end (import errors) so I just gave up and commented it  out for testing purposes. Will uncomment and investigate.

vulcanai2/models/basenetwork.py
nirtiac: ok sure I'll make that a parameter. 

Re torch.save - this is just a mvp pull request... but, torch.save seems to be a little complicated. if you look at the documentation [here](https://pytorch.org/tutorials/beginner/saving_loading_models.html) it says 

> The disadvantage of this approach is that the serialized data is bound to the specific classes and the exact directory structure used when the model is saved. The reason for this is because pickle does not save the model class itself. Rather, it saves a path to the file containing the class, which is used during load time.

and if you look at the code it does indeed do exactly that. so if we change anything, or heck, if people's directory structure is different, then we can no longer use what we saved. 

What I think should be done, but I need to investigate using different devices, is that we save the state_dict using torch.save, but then save our custom class instance using regular pickle. We then can reload the state dict appropriately for whichever device, and then manually, in our save function adjust everything else necessary for when transferring to a different device. I think this could wait for later, personally...

vulcanai2/models/basenetwork.py
```diff
@@ -492,12 +491,15 @@ def save_model(self, save_path=None):
         pickle.dump(self.state_dict, open(state_dict_file_path, "wb"), 2)  # TODO: pretty sure this isn't necessary
 
     # TODO: update the state dict and push to the appropriate device.
+    # TODO: caitrin save the optimizers state dict? even though this is included with our instance?
     # https://pytorch.org/tutorials/beginner/saving_loading_models.html#what-is-a-state-dict
+    # TODO: implement, add in classification and input layers?
     @classmethod
-    def load_ensemble(cls, load_path):
+    def load_model(cls, load_path, load_ensemble=True):
```
rfratila: perhaps `load_stacks` or `complete_model_stack` is a better name than `load_ensemble`? ensemble may be confused with Snapshot ensemble or other types of networks that are ensembles

PR #25, opened 2018-10-05
Saliency map via guided backpropagation implemented

rfratila:
Working on the tests now but want you guys to look it over while I do that.


priyatharsan:
I agree with adding `utils` folder.

vulcanai2/plotters/visualization.py
```diff
@@ -14,6 +14,8 @@
 
 from datetime import datetime
 
+from .utils import GuidedBackprop
+
```
priyatharsan: I don't see the `utils.py`, are you planning to add it as `utils/plotters.py`?

priyatharsan:
looks good to me 👍

vulcanai2/plotters/visualization.py
rfratila: yeah I plan on doing the folder restructure after I'm done

rfratila:
I found a weird bug with nested activations that have different names `activation` and `_activation`

priyatharsan:
I think @nirtiac removed them in the recent push?

nirtiac:
> it got me thinking, we should have a utils folder in our models, plotters, etc level so we have a consistent way of finding utility functions e.g. `from vulcanai.utils.plotters import *` instead of what we have now were we need to do `from vulcanai.plotters.utils import *`. Thoughts @priyatharsan @nirtiac ?

I don't see the issue? it's just the relevant model plus utils, it's just a different pattern and not more consistent, no?

vulcanai2/models/metrics.py
```diff
@@ -212,29 +144,29 @@ def run_test(self, model, test_x, test_y, figure_path=None, plot=False):
         print ('\nAccuracy: {}'.format(accuracy))
 
         print ('Sensitivity:'),
-        print(self.round_list(sens, decimals=3))
+        print(round_list(sens, decimals=3))
```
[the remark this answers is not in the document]
rfratila: yeah I think that's a good idea

PR #26, opened 2018-10-05
changed activations to _activations

priyatharsan:


priyatharsan:
the model is not saving although. Getting `OSError`
capture

rfratila:
Could you print model1 from the example script? I would like to see what the output looks like @priyatharsan

priyatharsan:
capture

priyatharsan:
Now it prints good enough 
capture

nirtiac:
@priyatharsan ok I somewhat expected that for the save function for windows... I've never used a windows machine so I don't know how it works. if it's an easy fix can you push it? if not I'll do some research

nirtiac:
@priyatharsan does the file name change just pushed help?

priyatharsan:
`saved_path` is now fixed, but I get "Python not responding" popup error during runtime of saving, but it gets saved.

examples/fashion_conv_dense_test.py
```diff
@@ -59,6 +59,7 @@
                         initializer=None,
                         bias_init=None, # None or value
                         norm=None,
+                        pool_size=None,
```
priyatharsan: good catch! I missed it

PR #29, opened 2018-10-10
MultiInputNN functionality update

priyatharsan:


vulcanai2/models/basenetwork.py
```diff
@@ -66,7 +66,14 @@ def __init__(self, name, dimensions, config, save_path=None, input_network=None,
 
         self._save_path = save_path
 
-        self._input_network = input_network
+        self._input_networks = input_networks
+        if isinstance(input_networks, list) and len(input_networks)==1:
+            self._input_network = input_networks[0]
```
rfratila: using both `self._input_networks` and `self._input_network` is confusing. you should just stick to using only `self._input_networks` so we have one point of entry for getting access to all incoming networks (even if it's just one)

vulcanai2/models/basenetwork.py
```diff
@@ -89,14 +96,46 @@ def __init__(self, name, dimensions, config, save_path=None, input_network=None,
             validation_accuracy=[]
         )
 
-        # self._itr = 0 #TODO: ?
+        self.in_dim = self._dimensions
+        if self._input_networks:
+            for i, in_net in enumerate(self._input_networks):
+                if in_net.__class__.__name__ == "ConvNet":
+                    if in_net.conv_flat_dim != self.in_dim:
+                        self.in_dim[i] = self.get_flattened_size(in_net)
```
rfratila: we shouldn't instantly flatten a ConvNet input, especially if it's incoming to another ConvNet object. Since this is in basenetwork, we should keep it very general and not specific to any inherited-child object (issues with circular object refs)

examples/fashion_conv_dense_test.py
```diff
@@ -101,14 +101,14 @@
...
-    input_network=None,
+    input_networks=None,
     dimensions=(1, 28, 28),
     config=conv_net_config,
 )
 
 model1 = DenseNet(
     name='dense_net_test',
-    input_network=model,
+    input_networks=[model],
```
rfratila: I thought we were using another dict method of passing models with options or would you rather leave it to the user to explicitly specify input network functionality after building the multi-input? e.g. model.freeze_input_networks()

vulcanai2/models/basenetwork.py
```diff
@@ -89,14 +96,46 @@ def __init__(self, name, dimensions, config, save_path=None, input_network=None,
...
+                        self.in_dim = in_net.dims[-1]
+                    else:
+                        pass
         self._create_network(
             activation=activation,
             pred_activation=pred_activation)
 
     # TODO: where to do typechecking... just let everything fail?
 
+    @abc.abstractmethod
```
rfratila: why do we have this if we already are using a `forward` function?

vulcanai2/models/basenetwork.py
```diff
@@ -89,14 +96,46 @@ def __init__(self, name, dimensions, config, save_path=None, input_network=None,
...
+    def _forward(self):
+        pass
+
+    def forward(self, inputs):
+        """Perform a forward pass through the module/modules."""
+
+        if self._input_networks:
```
rfratila: this check should be done much earlier (e.g.  __init__)

vulcanai2/models/basenetwork.py
```diff
@@ -89,14 +96,46 @@ def __init__(self, name, dimensions, config, save_path=None, input_network=None,
...
+            if not isinstance(inputs, list):
+                inputs = [inputs]
+            models=self._input_networks
+            
+            outputs = []
+            for model, x in zip(models, inputs):
+                outputs.append(model(x))
+            network_output = self._forward(torch.cat(outputs, 1))
```
rfratila: why are you using `forward`? Also please specify the function inputs explicitly so the code is more readable. (e.g. what does the 1 mean? probably the axis on which to concat but it's best to specify each attribute explicitly)

vulcanai2/models/basenetwork.py
```diff
@@ -199,8 +238,8 @@ def get_output_shapes(self, input_size=None):
         :return: OrderedDict of shape of each layer in the network
         """
         if not input_size:
-            if self._input_network:
-                input_size = self._input_network._dimensions
+            if self._input_networks:
+                input_size = self._input_networks._dimensions
```
rfratila: remember that you have a list for `_input_networks` so just calling `_dimesions` will not work. you will need to dynamically calculate it considering you concatenate the outputs. Also, we should have multiple checks and warning for non-matching incoming shapes (e.g. ConvNet and DenseNet incoming) so the user can decide the method for combining the two

vulcanai2/models/basenetwork.py
```diff
@@ -369,7 +408,7 @@ def _train_epoch(self, train_loader, retain_graph):
                 data, targets = data.cuda(), targets.cuda()
 
             # Forward + Backward + Optimize
-            predictions = self(data)
+            predictions = self([data])
```
rfratila: how are you handling multiple train_loaders being passed?

vulcanai2/models/basenetwork.py
```diff
@@ -414,7 +451,7 @@ def _validate(self, val_loader):
             if torch.cuda.is_available():
                 data, targets = data.cuda(), targets.cuda()
 
-            predictions = self(data)
+            predictions = self([data])
```
rfratila: same question about multiple val_loaders

vulcanai2/models/basenetwork.py
```diff
@@ -478,8 +515,13 @@ def save_model(self, save_path=None):
         os.makedirs(module_save_path)  # let this throw an error if it already exists
 
         # recursive recursive recursive
-        if self._input_network is not None:
-            self._input_network.save_model(save_path)
+        if self._input_networks is not None:
+            if self._input_network:
```
rfratila: you should probably loop through the `_input_networks` and save each of those first. We should revert to only using the `_input_networks` attribute instead of both that and `self_input_netowrk`

vulcanai2/models/cnn.py
```diff
@@ -99,17 +81,17 @@ def _create_network(self, **kwargs):
             self._create_classification_layer(
                 self.conv_flat_dim, kwargs['pred_activation'])
             
-            if torch.cuda.is_available():
-                for module in self.modules():
-                    module.cuda()
+        if torch.cuda.is_available():
```
rfratila: Do you need to do this for each module? can you not just apply it to `self.cuda()`?

vulcanai2/models/cnn.py
```diff
@@ -99,17 +81,17 @@ def _create_network(self, **kwargs):
...
+            for module in self.modules():
+                module.cuda()
 
     def _create_classification_layer(self, dim, pred_activation):
         self.network_tail = DenseUnit(
             dim, self.out_dim, activation=pred_activation)
 
-    def forward(self, x):
+    def _forward(self, x, **kwargs):
```
rfratila: same comment as above about turning forward into a hidden function. isn't `forward` required for any nn.Module object?

vulcanai2/models/dnn.py
```diff
@@ -122,11 +105,8 @@ def forward(self, x):
         :param x: input torch.Tensor
         :return: output torch.Tensor
         """
-        if self._input_network:
-            x = self._input_network(x)
 
-        if self._input_network and \
-            self._input_network.__class__.__name__ == "ConvNet":
+        if self._input_network and self._input_network.__class__.__name__ == "ConvNet":
```
rfratila: the backslash was done to shorten the length of that conditional for PEP8 reasons. why did you revert it?

vulcanai2/models/basenetwork.py
nirtiac: agreed

examples/fashion_conv_dense_test.py
nirtiac: there's also no need to do this; we could allow the user to pass multiple types and just use .isinstance() and then convert as necessary.

vulcanai2/models/basenetwork.py
nirtiac: actually, wait. that's what he did though.

vulcanai2/models/basenetwork.py
nirtiac: mmmm, only if we don't want to allow people to set input networks after the object is initialized. do we? thoughts?

vulcanai2/models/basenetwork.py
```diff
@@ -61,18 +61,21 @@ def __init__(self, name, dimensions, config, save_path=None, input_networks=None
...
         self._config = config
 
         self._save_path = save_path
 
-        self._input_networks = input_networks
-        if isinstance(input_networks, list) and len(input_networks)==1:
-            self._input_network = input_networks[0]
-        elif isinstance(input_networks, type(None)):
-            self._input_network = None
+        if not isinstance(input_networks, type(None)) and \
```
rfratila: why check if it's type(None)?

vulcanai2/models/basenetwork.py
rfratila: yeah I think we should be able to add an input network after it's been initialized, or add more

vulcanai2/models/basenetwork.py
```diff
@@ -131,10 +140,15 @@ def forward(self, inputs):
             outputs = []
             for model, x in zip(models, inputs):
                 outputs.append(model(x))
```
rfratila: need to test if each output shape gets independently preserved in the outputs list

vulcanai2/models/basenetwork.py
```diff
@@ -278,9 +289,15 @@ def hook(module, input, output):
                     not (module == self):
                 hooks.append(module.register_forward_hook(hook))
 
+        # check if input_size is not a int
+        if isinstance(input_size, int):
+            input_size = [input_size]
+        # check if there is only one input to the network
+        if isinstance(input_size[0], (list, tuple)) and len(input_size)==1:
+            x = Variable(torch.rand(1, *input_size[0]))
```
rfratila: you can just use torch.empty for this, saves memory

vulcanai2/models/dnn.py
```diff
@@ -106,16 +106,11 @@ def _forward(self, x, **kwargs):
         :return: output torch.Tensor
         """
 
-        if self._input_network and self._input_network.__class__.__name__ == "ConvNet":
-            x = x.view(-1, self._input_network.conv_flat_dim)
-
+        if self._input_networks and self._input_networks[0].__class__.__name__ == "ConvNet":
```
rfratila: you should check for more than just the first input network. need to go through all of them

vulcanai2/models/basenetwork.py
```diff
@@ -96,33 +99,39 @@ def __init__(self, name, dimensions, config, save_path=None, input_networks=None
...
     # TODO: where to do typechecking... just let everything fail?
 
-    @abc.abstractmethod
-    def _forward(self):
-        pass
-
     def forward(self, inputs):
         """Perform a forward pass through the module/modules."""
-
         if self._input_networks:
```
rfratila: check if it's not None

vulcanai2/models/basenetwork.py
```diff
@@ -99,50 +103,50 @@ def __init__(self, name, dimensions, config, save_path=None, input_networks=None
...
-            
             outputs = []
             for model, x in zip(models, inputs):
                 outputs.append(model(x))
-            network_output = self._forward(torch.cat(outputs))
+                #TODO: Use tablemodules NEW: https://github.com/torch/nn/blob/master/doc/table.md
+            network_output = self._forward(torch.s(outputs, dim=1)) # Ref:https://github.com/torch/torch7/blob/master/doc/maths.md
         else:
-            network_output = self._forward((inputs))
+            network_output = self._forward((inputs[0]))
```
rfratila: you don't need to do this if only one model is passed. just treat it as if there are many passed

vulcanai2/models/basenetwork.py
```diff
@@ -215,22 +219,22 @@ def criter_spec(self, value):
...
     def get_flattened_size(self, network):
         """
-        Returns the flattened output size of the conv network's last layer.
+        Returns the flattened output size of a Single Input ConvNet's last layer.
         :param network: The network to flatten
         :return: The flattened output size of the conv network's last layer.
         """
         with torch.no_grad():
-            x = torch.ones(1, *self.in_dim)
+            x = torch.ones(1, *self.in_dim[0])
```
rfratila: there should be no index-based retrievals. what if the conv wasn't in the first slot?

vulcanai2/models/basenetwork.py
```diff
@@ -215,22 +219,22 @@ def criter_spec(self, value):
...
 
     def get_output_size(self):
         """
-        Returns the output size of the network's last layer
+        Returns the output size of the Single Input DenseNet's last layer
         :return: The output size of the network's last layer
         """
         with torch.no_grad():
-            x = torch.ones(1, self.in_dim)
+            x = torch.ones(1, self.in_dim[0])
```
rfratila: same comment as above on index-based retrieval

vulcanai2/models/basenetwork.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, name, dimensions, config, save_path=None, input_networks=None
         """
         Defines the network object.
         :param name: The name of the network. Used when saving the file.
-        :param dimensions: The dimensions of the network. type: List of tuples
+        :param dimensions: The dimensions of the network. type: List of tuples or int
```
rfratila: since you removed _dimensions, we should rename this to `in_dims` for consistency variable naming

vulcanai2/models/basenetwork.py
```diff
@@ -61,14 +61,18 @@ def __init__(self, name, dimensions, config, save_path=None, input_networks=None
...
 
         self._name = name
-        if isinstance(dimensions, tuple):
-            self._dimensions = [dimensions]
-        elif isinstance(dimensions, int):
-            self._dimensions = (dimensions,)
+        if dimensions is not None:
+            # Must be a list of tuples(ConvNet)/int(DenseNet)
+            if isinstance(dimensions, (tuple, int)):
+                self.in_dim = [dimensions]
```
rfratila: why wrap it in a list if it is already a tuple?

vulcanai2/models/basenetwork.py
```diff
@@ -61,14 +61,18 @@ def __init__(self, name, dimensions, config, save_path=None, input_networks=None
...
+            # Must be a list of tuples(ConvNet)/int(DenseNet)
+            if isinstance(dimensions, (tuple, int)):
+                self.in_dim = [dimensions] 
+            else:
+                self.in_dim = dimensions
         else:
-            self._dimensions = dimensions
+            self.in_dim = []
+            for net in self._input_networks:
+                self.in_dim.append(net.in_dim)
```
rfratila: why do you care about the in_dim of the input_networks? only the output shape should matter no? with regards to .in_dim

vulcanai2/models/basenetwork.py
```diff
@@ -99,50 +103,50 @@ def __init__(self, name, dimensions, config, save_path=None, input_networks=None
...
-                    
-    #                 self.in_dim.append(net.get_output_shapes())
-
-    #             else:
-    #                 self.in_dim = self.get_in_dim(net)
-    #             temp_tensor = torch.zeros([1, *net.in_dim])
+
+        #print(self)
+        
+        out_shapes = self.get_output_shapes(network=self.network,input_size=self.in_dim)
```
rfratila: if it's already a class function, you shouldn't need to pass it a self object

vulcanai2/models/basenetwork.py
```diff
@@ -246,13 +250,16 @@ def get_size(self, summary_dict, output):
...
+    def get_output_shapes(self, network=None, input_size=None):
         """
         Returns the summary of shapes of all layers in the network
         :return: OrderedDict of shape of each layer in the network
         """
         if not input_size:
-            input_size = self._dimensions
+            input_size = self.in_dim
+        # input_size must be a list
+        if isinstance(input_size, (tuple, int)):
```
rfratila: why create a list of a tuple?

vulcanai2/models/basenetwork.py
priyatharsan: I think I mentioned in the documentation "works only for Single Input case"

vulcanai2/models/basenetwork.py
priyatharsan: Same - "works only for Single Input case"

vulcanai2/models/basenetwork.py
rfratila: why not make it work for all input cases?

vulcanai2/models/basenetwork.py
```diff
@@ -498,6 +499,20 @@ def _validate(self, val_loader):
         validation_accuracy = val_accuracy_accumulator * len(data) / len(val_loader.dataset)
 
         return validation_loss, validation_accuracy
+
```
rfratila: I think when you apply cuda to a model, it automatically applies it to incoming models

vulcanai2/models/basenetwork.py
priyatharsan: didn't we agree on list of tuples?

vulcanai2/models/basenetwork.py
priyatharsan: I'm overriding it to just do a forward pass on self.network alone. Check here: https://github.com/Aifred-Health/Vulcan2/blob/update_branch2/vulcanai2/models/basenetwork.py#L310-L320

vulcanai2/models/basenetwork.py
priyatharsan: I think it is more protected  like that, how do you suggest to handle a networks in_dim containing image data, text data and tabular data?
`in_dim=[(conv_in_dim), dense_in_dim] = [(1,28,28), 784]`? Not sure what could be a better alternative

vulcanai2/models/basenetwork.py
rfratila: I see. we will need to come back to get_output_shapes to clean it up.

vulcanai2/models/basenetwork.py
rfratila: you're right. let's just make sure that we have a list of tuples, even for dense_in_dim. e.g. `[(1,28,28), (784)]` so when we interface with it, the type is consistent of the inputs

priyatharsan: I tried to have the same for dense with `(784)` it reads as a int, it should of been like `(784,)` with a comma, but this implementation is complicating the inputs to the model

vulcanai2/models/basenetwork.py
priyatharsan: NopeI get this `RuntimeError: Input type (torch.cuda.FloatTensor) and weight type (torch.FloatTensor) should be the same`
Thats because our model workflow involves lists of model?. I think so.

vulcanai2/models/basenetwork.py
priyatharsan: `MultiinputNN.in_dim` should have a list of `in_dims`? No?

priyatharsan: btw it `in_dim` not with a 's'

vulcanai2/models/basenetwork.py
priyatharsan: I wanted it to be a helper function to get the flattened size off a single input network, I think for multpple case, we can use get_output_shapes?

vulcanai2/models/basenetwork.py
rfratila: can't you cast it to list with `list(784)`?

vulcanai2/models/basenetwork.py
priyatharsan: yeah that was dumb lol

vulcanai2/models/basenetwork.py
rfratila: we shouldn't have a conv specific function in basenetwork

vulcanai2/models/basenetwork.py
rfratila: then we need to update those areas because we can't have an in_dim=[(1,28,28), 256, (3, 200, 200, 200)] bc each time we go through in_dim we need to check the types instead of just assuming they are all the same. does that make sense?

priyatharsan: Also, let's say we have an input structure like [conv, dense, dense, conv], so having in_dim like `[(1,28,28), 784, 500, (6, 25, 25)]` will make more sense than, `[(1,28,28), [784, 500], (6, 25, 25)]`, as we use only one for loop to point to the models in self.forward which suits the first case. Does that makes sense?

rfratila: no it should be `[(1,28,28), [784], [500], (6, 25, 25)]`. each in their own list

priyatharsan: I'm not sure how I can revert back `(784), (500),`to `tuple` form when it gets read as `int`. Doing like  `[784], [500]` will eliminate the purpose of type checking for list of inputs (atleast until if we decide to move to dict type inputs). And as long it is `[(1,28,28), (784), (500), (6, 25, 25)]` or `[(1,28,28), 784, 500, (6, 25, 25)]` `(tuple/int)` it is going to considered as a separate `in_dim`, that's what the type checking is doing there

vulcanai2/models/cnn.py
```diff
@@ -92,8 +94,42 @@ def __init__(self, name, in_dim, config, save_path=None,
             lr_scheduler, early_stopping, criter_spec)
 
     def _create_network(self, **kwargs):
-
+        self._in_dim = self.in_dim
```
rfratila: Why do we have two of the same variables?

examples/fashion_conv_dense_test.py
```diff
@@ -107,7 +107,7 @@
 model = ConvNet(
     name='conv_net_test',
     input_networks=None,
-    in_dim=(1, 28, 28),
+    in_dim=[(1, 28, 28)],
```
rfratila: wrapping it in a list should be done behind the scenes

vulcanai2/models/cnn.py
```diff
@@ -92,8 +94,42 @@ def __init__(self, name, in_dim, config, save_path=None,
...
         conv_hid_layers = self._config.units
+
+        if len(self.in_dim) > 1:
```
rfratila: you have to be very careful with this conditional. If for some reason it's not a list of tuples than a simple convnet with dim (1, 28, 28) will think it's 3 separate inputs

vulcanai2/models/cnn.py
```diff
@@ -92,8 +94,42 @@ def __init__(self, name, in_dim, config, save_path=None,
...
+            # Calculate converged in_dim for the MultiInput ConvNet
+            dim_tmp = []
+            # Create empty input tensors
+            in_tensors = [torch.empty(x) for x in self.in_dim]
+            # Calculate the number of elements in each input tensor 
+            els = [t.numel() for t in in_tensors]
```
rfratila: this will take into account the batch size as well. Are you sure numel is what you want to use?

vulcanai2/models/cnn.py
```diff
@@ -92,8 +94,42 @@ def __init__(self, name, in_dim, config, save_path=None,
...
+            #Calculate the size of each input tensor
+            dim_sizes = [len(t.size()) for t in in_tensors]
```
rfratila: Don't you already know this from `.in_dim`?

vulcanai2/models/cnn.py
```diff
@@ -92,8 +94,42 @@ def __init__(self, name, in_dim, config, save_path=None,
...
+
+            max_el_ind = els.index(max(els))
```
rfratila: Can you explain what this does?

vulcanai2/models/cnn.py
```diff
@@ -92,8 +94,42 @@ def __init__(self, name, in_dim, config, save_path=None,
...
+            max_dim_ind = dim_sizes.index(max(dim_sizes))
+
+            if max_el_ind == max_dim_ind:
```
rfratila: This needs to be much more thoroughly documented since a lot of complicated stuff is happening.

vulcanai2/models/cnn.py
```diff
@@ -161,21 +159,50 @@ def _build_conv_network(self, conv_hid_layers, activation):
...
+        xs : list(torch.Tensor)
+            List of input tensors to pass through self.
+
+        Returns
+        -------
+        output : torch.Tensor
+
+        """
+        out = []
+        for x in xs:
```
rfratila: What is the point of this? you're not running it through any network. might as well just cat `xs` no?

vulcanai2/models/utils.py
```diff
@@ -111,4 +111,15 @@ def get_size(summary_dict, output):
             summary_dict[i] = get_size(summary_dict[i], output[i])
     else:
         summary_dict['output_shape'] = list(output.size())
-    return summary_dict
\ No newline at end of file
+    return summary_dict
+
+def expand_dim(tensor, max_dim=None, axis=0):
```
rfratila: Please provide documentation for this.

examples/fashion_conv_dense_test.py
priyatharsan: Yep, it was just for testing purpose, the Basenetwork internally handles them as list of tuples/int (they get converted internally), which we may revisit once we have functional MultiInputNN.

vulcanai2/models/cnn.py
priyatharsan: yeah we are using `torch.cat` in the end. the input tensors needed to be formatted before passing into the network, it will be implemented soon.

vulcanai2/models/cnn.py
priyatharsan: this is essential to know for everyone, maybe we might change the usage in future, but as of now,
self.in_dim spits out dimensions of whatever input passed into a network, while self._in_dim (private) provides the dimension size of the actual input that get fed into the network
For example: if `self.in_dim = [(1,28,28), 784, (64, 2, 2)]`, then `self._in_dim = [(66, 28, 28)]`

vulcanai2/datasets/multidataset.py
```diff
@@ -0,0 +1,99 @@
...
+        (input_data, targets) : (torch.Tensor, torch.Tensor)
+            Tuple of input_data and target at that index as specific in config.
+            Target data is always last.
+
+        """
+        input_data_items = []
+        target_items = []
+
+        for t in self._datasets:
+            if isinstance(t, MultiDataset):
```
rfratila: @nirtiac can you help think about what the best way to return list of lists for input_data but not for targets? this is the last piece of the puzzle

vulcanai2/datasets/multidataset.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+
+    # TODO: is datasets a reasonable name?
+    # TODO: would be better to make these namedtuples...
+    def __init__(self, datasets):
+        """Initialize a dataset for multi input networks."""
+        def get_total_targets(multi_datasets):
+            num_targets = 0
+            for ds in multi_datasets:
+                if isinstance(ds, MultiDataset):
```
priyatharsan: shouldn't this be checking for both object types? `MultiDataset` and `Dataset`

rfratila: no because we only need to recurse if it's a multi-dataset. all other Datasets will be accounted for in the else case @priyatharsan

rfratila: try changing the target boolean from the nested one to the top layer multi-dataset in `fashion_....py` and test that it returns the correct target

PR #32, opened 2018-10-11
Saliency fix, gradient error propagation fix, visualization, documentation and code style fix

rfratila:
- Removed double softmax transformations
  - Moved  to only `forward_pass` since criterion already applies it for training
- Removed default activation for `pred_activation` so we don't do Softmax(Relu(x)) during training. Left the variable for if it's an intermediate layer
- Added more tests

### DIFF
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -434,11 +434,15 @@ class BaseNetwork(nn.Module):
 
         return validation_loss, validation_accuracy
 
-    def run_test(self, test_x, test_y, figure_path=None, plot=False):
+    def run_test(self, data_loader, figure_path=None, plot=False):
         """
         Will conduct the test suite to determine model strength.
         """
-        return self.metrics.run_test(self, test_x, test_y, figure_path, plot)
+        return self.metrics.run_test(
+            model=self,
+            data_loader=data_loader,
+            figure_path=figure_path,
+            plot=plot)
 
     # TODO: Instead of self.cpu(), use is_cuda to know if you can use gpu
     def forward_pass(self, data_loader, convert_to_class=False):
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -439,7 +439,7 @@ class BaseNetwork(nn.Module):
         Will conduct the test suite to determine model strength.
         """
         return self.metrics.run_test(
-            model=self,
+            network=self,
             data_loader=data_loader,
             figure_path=figure_path,
             plot=plot)
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -75,8 +75,7 @@ class BaseNetwork(nn.Module):
         self._criter_spec = criter_spec
 
         if self._num_classes:
-            self.criter_spec = nn.CrossEntropyLoss()
-            pred_activation = nn.ReLU()
+            self._criter_spec = nn.CrossEntropyLoss()
             self.metrics = Metrics(self._num_classes)
 
         self.optim = None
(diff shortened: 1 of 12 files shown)


vulcanai2/models/basenetwork.py
```diff
@@ -453,11 +454,16 @@ def forward_pass(self, input_data, convert_to_class=False):
         Returns: Numpy matrix with the output probabilities
                  with each class unless otherwise specified.
         """
-        output = self.cpu()(torch.Tensor(input_data)).data.numpy()
-        if convert_to_class:
-            return self.metrics.get_class(output)
-        else:
-            return output
+        self.eval()
+        if not isinstance(input_data, torch.Tensor):
```
nirtiac: input data should be the same as fit, aka take a data_loader and perform necessary operations to get the input_data. Also add a parameter for number of samples to evaluate on.

vulcanai2/plotters/utils.py
```diff
@@ -64,21 +71,28 @@ def remove_hooks(self):
         for h in self.hooks:
             h.remove()
 
-    def generate_gradients(self, input_x, target_class_list):
+    def generate_gradients(self, input_x, targets):
```
nirtiac: this should also take the same input type... and we should be consistent in our naming of variables and parameters in regards to data/x/training and targets/y data.

vulcanai2/models/cnn.py
```diff
@@ -105,7 +105,9 @@ def _create_network(self, **kwargs):
 
     def _create_classification_layer(self, dim, pred_activation):
         self.network_tail = DenseUnit(
-            dim, self.out_dim, activation=pred_activation)
+            in_features=dim,
+            out_features=self.out_dim,
+            activation=pred_activation)
```
nirtiac: TODO: check for correct dimensions between input data and self.dimensions. Also remove self.in_dim from _create_network

vulcanai2/models/metrics.py
```diff
@@ -78,6 +80,8 @@ def get_class(self, in_matrix):
         :param n_matrix: one-hot matrix
         :return: 1D Class array
         """
+        if isinstance(in_matrix, torch.Tensor):
+            in_matrix = in_matrix.detach().numpy()
```
priyatharsan: Getting an error at this point after commenting out `model1.runtest` in the example
```
Traceback (most recent call last):
  File "fashion_conv_dense_test.py", line 136, in <module>
    model1.forward_pass(val_loader, convert_to_class=True),
  File "c:\work\aifred\vulcan2\master\vulcan2\vulcanai2\models\basenetwork.py", line 473, in forward_pass
    self.metrics.get_class(in_matrix=predictions)).float()
  File "c:\work\aifred\vulcan2\master\vulcan2\vulcanai2\models\metrics.py", line 84, in get_class
    in_matrix = in_matrix.detach().numpy()
TypeError: can't convert CUDA tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```
Same CUDA issue.

vulcanai2/plotters/utils.py
```diff
@@ -22,66 +23,81 @@ def __init__(self, network):
...
+        self.network.eval()
+        # To properly pass the gradients
+        if not isinstance(input_data, torch.Tensor):
+            input_data = torch.tensor(input_data, requires_grad=True)
+        else:
+            if not input_data.requires_grad:
+                input_data.requires_grad = True
         # Forward pass
-        network_output = self.network(input_x)
+        network_output = self.network(input_data)
```
priyatharsan: Change this to: `network_output = self.network.cpu()(input_data)`

vulcanai2/models/basenetwork.py
```diff
@@ -434,30 +434,51 @@ def _validate(self, val_loader):
...
+            # Get raw network output
+            predictions = self(data)
+            if self._num_classes:
+                # Get probabilities
+                predictions = nn.Softmax(dim=1)(predictions)
+                if convert_to_class:
+                    predictions = torch.tensor(
+                        self.metrics.get_class(in_matrix=predictions)).float()
+            # Aggregate predictions
+            pred_collector = torch.cat([pred_collector, predictions])
```
priyatharsan: Change to: `pred_collector = torch.cat([pred_collector, predictions.cpu()])`

vulcanai2/models/metrics.py
priyatharsan: Actually change it to: `in_matrix = in_matrix.cpu().detach().numpy()`

rfratila: im just gonna change it before it gets called into get_class

vulcanai2/plotters/utils.py
rfratila: will do

vulcanai2/models/basenetwork.py
rfratila: will do

PR #34, opened 2018-10-12
Dataset Wrapper for Multi-Input Use

nirtiac:
Still needs to be tested but the concept is there.


rfratila:
@nirtiac is it possible to do multi-datasets of multi-datasets?

PR #35, opened 2018-10-12
Cross validation

nirtiac:
Needs rebasing and tests


nirtiac:
@priyatharsan and @rfratila please review!

vulcanai2/models/metrics.py
```diff
@@ -152,11 +151,12 @@ def run_test(self, network, data_loader, figure_path=None, plot=False):
            network._num_classes == 0:
             raise ValueError('There\'s no classification layer')
 
-        test_y = data_loader.dataset.test_labels
+        test_y = np.array([v[1] for v in data_loader.dataset])
```
rfratila: can you add some docs here about what this is doing?

vulcanai2/models/metrics.py
```diff
@@ -241,73 +241,83 @@ def run_test(self, network, data_loader, figure_path=None, plot=False):
...
+        from torch.utils.data import TensorDataset
+
+        all_results = defaultdict(lambda: [])
+
+        # TODO: this whole section is really clunky
+        fold_len = math.floor(data_loader.dataset.__len__() / k)
+        rem = data_loader.dataset.__len__() % k
+        fold_seq = []
+
+        print(fold_seq)
```
rfratila: perhaps we don't need to print this? or just use a logger?

vulcanai2/tests/models/test_metrics.py
```diff
@@ -2,13 +2,42 @@
...
+from vulcanai2.models.cnn import ConvNet
+from vulcanai2.models.dnn import DenseNet
+from vulcanai2.models.ensemble import SnapshotNet
+from torch.utils.data import TensorDataset, DataLoader
 
 class TestMetrics:
     @pytest.fixture
     def metrics(self):
         return Metrics(
             num_class=2
```
rfratila: you have num_classes as 2 here and 6 when you make the network? we might as well remove it from metrics no?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -56,8 +57,11 @@ def __getitem__(self, idx):
         """
         # Where df.drop is used to access the dataframe without the label column, iloc gets the row, then access values
         # and convert
+
         xs = self.df.drop(self.labelColumn, axis=1).iloc[[2]].values.tolist()[0]
-        y = self.df[[self.labelColumn]].iloc[[idx]].values.tolist()[0]
+        xs = torch.tensor(xs, dtype=torch.float)
+        y = self.df[[self.labelColumn]].iloc[[idx]].values.tolist()[0][0]
+        y = torch.tensor(y, dtype=torch.long)
```
rfratila: would we be able to get the type of y and use that to inform if we should use `dtype=torch.float` or `dtype=torch.long` @joemehltretter ?

PR #54, opened 2018-10-19
Freeze and unfreeze networks implemented

rfratila:


rfratila:
Actually would it make more sense to set requires_grad= ~freeze? we wouldn't need all the conditional statements but it's harder to read. thoughts @joemehltretter @nirtiac @priyatharsan ?

PR #56, opened 2018-10-20
Fix circular imports

rfratila:
I apologize for the overlap with the freeze weights PR (not exactly sure why it did that but anyway). I just moved the get_notable_indices to the utils for plotters. you can now have it like we did before with `from vulcanai2.models import DenseNet` instead of `from vulcanai2.models.dnn import DenseNet`

### DIFF
--- a/vulcanai2/plotters/utils.py
+++ b/vulcanai2/plotters/utils.py
@@ -4,6 +4,31 @@ from torch.nn import ReLU, SELU
 # from ..models.basenetwork import BaseNetwork
 
 
+def get_notable_indices(feature_importances, top_k=5):
+    """
+    Return dict of top k and bottom k features useful from matrix.
+
+    Parameters
+    ----------
+    feature_importance : numpy.ndarray
+        1D numpy array to extract the top of bottom indices.
+    top_k : int
+        How many features from top and bottom to extract.
+        Defaults to 5.
+
+    Returns
+    -------
+    notable_indices : dict
+        Indices of the top most important features.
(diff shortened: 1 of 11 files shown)


rfratila:
Actually would it make more sense to set `requires_grad= ~freeze`? we wouldn't need all the conditional statements but it's harder to read. thoughts @joemehltretter @nirtiac @priyatharsan ?

vulcanai2/plotters/utils.py
```diff
@@ -38,10 +63,10 @@ def __init__(self, network):
     def _hook_top_layers(self):
         def hook_function(module, grad_in, grad_out):
             # TODO: Revisit dim disorder and check isinstance for classes.
-            if module.__class__.__name__ == 'Linear':
+            if isinstance(module, torch.nn.Linear):
                 # grad_in shape is (bias, input, weights)
                 self.gradients = grad_in[1]
-            elif module.__class__.__bases__[0].__name__ == '_ConvNd':
+            elif isinstance(module, torch.nn.modules.conv._ConvNd):
```
priyatharsan: Just curious, is there any diff wrt to performance by doing this? instead of checking with base Class name?

nirtiac: my impression was because it lets you deal with complex class hierarchies

priyatharsan: I don't get it. What do you mean?

priyatharsan: 👍  thanks

PR #58, opened 2018-10-22
Update stitch dataset

joemehltretter:
Did additional testing and functions properly. I removed `fillna()` from the function because it makes the `remove_majority_null()` function in _tabulardataset.py_ difficult to use. I think we can create a seperate `fillna` function in _tabulardataset.py_ that the user can control. This I think will be more useful as the user can control when and how to fill null values. I added print functions just so the user knows the function is working, as it can take a little time depending on the number and size of dataframes and the number of keys to join on.

### DIFF
--- a/vulcanai2/datasets/utils.py
+++ b/vulcanai2/datasets/utils.py
@@ -161,8 +161,8 @@ def stitch_datasets(df_list, on, index_list=None):
 
         merged_df = df_groupOn.drop_duplicates(subset=on)
 
-    merged_df.fillna(-1, inplace=True)
     print("\nMerge Total columns = {totalCols}, rows = {totalRows} ".format(
         totalCols=len(list(merged_df)),
         totalRows=len(merged_df)))
     return merged_df
+
--- a/vulcanai2/datasets/utils.py
+++ b/vulcanai2/datasets/utils.py
@@ -7,6 +7,7 @@ for anaconda and is therefore not yet a reasonable dependency.
 See https://github.com/pytorch/text/blob/master/torchtext/data/dataset.py
 """
 import pandas as pd
+import random
 import logging
 import copy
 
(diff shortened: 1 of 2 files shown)


vulcanai2/datasets/utils.py
```diff
@@ -136,55 +136,33 @@ def stitch_datasets(df_list, on, index_list=None):
...
-    merged_df.fillna(0, inplace=True)
-
+    first_column = list(df_list)[1]
+    merged_df = df_list[first_column].copy(deep=True)
+    merged_df = merged_df.apply(pd.to_numeric, errors='ignore')
+
+    count = 1
+
+    for key in list(df_list):
+        if key != first_column:
```
rfratila: once you get past the first column you don't really need to keep checking. Perhaps we can just updated what key in df_list to loop through by omitting that key?

vulcanai2/datasets/utils.py
```diff
@@ -136,55 +136,33 @@ def stitch_datasets(df_list, on, index_list=None):
...
+            print('Combining: {} '.format(key))
+
+            count = count + 1
+            df_two = df_list[key].copy(deep=True)
+            df_two = df_two.apply(pd.to_numeric, errors='ignore')
+            merged_df = merged_df.append(df_two)
+
+    if on is not None:
```
rfratila: Let's come up with a more informative var name than `on`

vulcanai2/datasets/utils.py
```diff
@@ -136,55 +136,33 @@ def stitch_datasets(df_list, on, index_list=None):
...
+        # Group by keys, forward fill and backward fill missing data then remove duplicate keys
+        df_groupOn = merged_df.reset_index().groupby(on).apply(
+            lambda x: x.fillna(method='ffill').fillna(method='bfill'))
+        if 'index' in list(df_groupOn):
+            del df_groupOn['index']
+        print("\tDropping duplicates")
```
rfratila: let's use loggers instead of print statements

rfratila:
Would you also be able to include some unit tests?

vulcanai2/datasets/utils.py
```diff
@@ -136,55 +136,33 @@ def stitch_datasets(df_list, on, index_list=None):
...
-                concat_list[1::2] = col_list_1
-                concat_list[2::2] = col_list_2
-                df_list[i] = df_list[i][concat_list]
-
-    for j in range(len(df_list)):
-        merged_df = pd.merge(merged_df, df_list[j], how='outer', on=on)
-
-    merged_df.fillna(0, inplace=True)
-
+    first_column = list(df_list)[1]
```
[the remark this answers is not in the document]
joemehltretter: It just grabs the name of the first dataframe we'll merge. That way it can be referenced so we don't double over it when iterating through through the dataframes to merge.

vulcanai2/datasets/utils.py
joemehltretter: We can remove the dataframe. I'll update that.

vulcanai2/datasets/utils.py
joemehltretter: would `merge_on` or `key` be better?

vulcanai2/datasets/utils.py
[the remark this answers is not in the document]
rfratila: ok let's make a comment about that since it's not immediately obvious

vulcanai2/datasets/utils.py
rfratila: great

vulcanai2/datasets/utils.py
rfratila: `merge_on_column`?

vulcanai2/datasets/utils.py
nirtiac: here you would just go logger.info("\tDropping duplicates")

nirtiac:
@rfratila to be fair to @joemehltretter  it would probably be much easier to add tests for this once I've finished the other dataset tests, which I'm still muddling through

rfratila:
Fair enough @nirtiac . We'll leave that for later.

joemehltretter:
@rfratila @nirtiac I began the unit test in the test_utils.py file. Is that okay or should I create a new file?

rfratila:
I'll leave it up to @nirtiac 's judgement but as long as it's in the file with all the other tabular dataset tests it's fine with me

rfratila:
go for it!

PR #77, opened 2018-11-02
Fix #43 - Remove network_tail. Added names to layers.

### LINKED ISSUE #43: do we need to have a dedicated class attribute for network_tail? why not just add it to network?

rfratila:


vulcanai2/models/cnn.py
```diff
@@ -271,20 +273,22 @@ def _build_conv_network(self, conv_hid_layers, activation):
...
         """
         conv_hid_layers[0]['in_channels'] = self.in_dim[0]
-        conv_layers = []
-        for conv_layer_config in conv_hid_layers:
+        conv_layers = OrderedDict()
+        for idx, conv_layer_config in enumerate(conv_hid_layers):
             conv_layer_config['activation'] = activation
-            conv_layers.append(ConvUnit(**conv_layer_config))
-        conv_network = nn.Sequential(*conv_layers)
+            layer_name = 'conv_{}'.format(idx)
```
priyatharsan: Can we generate a random number as ID for the network, say `network_ID`, and link to the layer names? I know it will increase the verbosity level but I feel it will be useful for complex multi-input networks. Thoughts?

vulcanai2/models/cnn.py
```diff
@@ -271,20 +273,22 @@ def _build_conv_network(self, conv_hid_layers, activation):
...
-                    in_features=dim,
-                    out_features=self._num_classes,
-                    activation=pred_activation))
+        self.network.add_module(
+            'flatten', FlattenUnit())
+        self.network.add_module(
+            'classify', DenseUnit(
+                in_features=dim,
+                out_features=self._num_classes,
+                activation=pred_activation))
```
priyatharsan: I would suggest  doing this within `_build_network` itself

vulcanai2/models/dnn.py
```diff
@@ -144,12 +146,13 @@ def _create_network(self, **kwargs):
...
-                    in_features=dim,
-                    out_features=self._num_classes,
-                    activation=pred_activation))
+        self.network.add_module(
+            'flatten', FlattenUnit())
+        self.network.add_module(
+            'classify', DenseUnit(
+                in_features=dim,
+                out_features=self._num_classes,
+                activation=pred_activation))
```
priyatharsan: same: I would suggest doing this within _build_network itself

vulcanai2/models/cnn.py
rfratila: well we already have the network name as a unique Identifier so we won't need to generate another ID. I will try to add it to when we assign the input_networks so as to tie the input names to the names of the input networks. 
screen shot 2018-11-02 at 3 43 38 pm

For instance here, instead of `0` it can say 'Conv_1D_name'. What do you say @priyatharsan ?

priyatharsan: Yeah I agree on that. I was actually suggesting to put more info like network_ID_layer_name (e.g.: `net1234_conv1D_1`, `net1234_flatten`, `net1234_classifier`). I'm good with just the layer_name also.

priyatharsan:
>I think that I can just move _create_classification_layer to BaseNetwork. thoughts?
I just saw this. I'm good with it.

vulcanai2/models/dnn.py
rfratila: the issue is that get_flattened_size requires a `self.network` to already exist @priyatharsan

priyatharsan: oh dont worry about that. At this moment lets make it optional to compute flattened size only if the network does not have a `network_tail`. We were using flattened_size for shape inferencing between `ConvNet` and `DenseNet` outside of the `BaseNetwork` container, remember? Now it is nearly not required to have that computation.

rfratila: if we don't know the flattened size then how would you set up your last DenseUnit? How else would you determine the in_features?

priyatharsan: Ah I see you are right! How about we make it like this:

```

    def _create_network(self, **kwargs):
        conv_hid_layers = self._config.units

        if self.input_networks is not None:
            # Create empty input tensors
            in_tensors = []
            for d in self.input_networks:
                in_tensors.append(torch.ones([1, *d.out_dim]))
            output = self._merge_input_network_outputs(in_tensors)
            self.in_dim = tuple(output.shape[1:])

        # Build Network
        self.network = self._build_conv_network(
            conv_hid_layers, kwargs['activation'])


    def _build_conv_network(self, conv_hid_layers, activation):
        """
        Build the layers of the network into a nn.Sequential object.
        Parameters
        ----------
        conv_hid_layers : ConvNetConfig.units (list of dict)
            The hidden layers specification
        activation : torch.nn.Module
            the non-linear activation to apply to each layer
        Returns
        -------
        output : torch.nn.Sequential
            the conv network as a nn.Sequential object
        """
        conv_hid_layers[0]['in_channels'] = self.in_dim[0]
        conv_layers = OrderedDict()
        for idx, conv_layer_config in enumerate(conv_hid_layers):
            conv_layer_config['activation'] = activation
            layer_name = 'conv_{}'.format(idx)
            conv_layers[layer_name] = ConvUnit(**conv_layer_config)
        conv_network = nn.Sequential(conv_layers)

        if self._num_classes:
           conv_network .add_module(
                   'flatten', FlattenUnit())
           conv_network .add_module(
                'classify', DenseUnit(
                    in_features=self.get_flattened_size(),
                    out_features=self._num_classes,
                    activation=pred_activation))

       return conv_network

Comment (rfratila) [reply vulcanai2/models/dnn.py]: yeah that's what I tried but it will crash as soon as you run `self.get_flattened_size()`. self.network doesn't yet exist

File: vulcanai2/models/cnn.py (hunk 1 above)
Comment (rfratila) [reply vulcanai2/models/cnn.py]: the name would be a little too repetitive having `net1234` at each layer. if we can attach it at the  ConvNet level that would be the cleanest implementation?

File: vulcanai2/models/dnn.py (hunk 3 above)
Comment (priyatharsan) [reply vulcanai2/models/dnn.py]: Yeah I did a slight modification now ⬆️

Comment (priyatharsan) [reply vulcanai2/models/dnn.py]: whats more funny is you can further replace `in_features=self.get_flattened_size()` with `in_features=self._get_out_dim()[0]` 😁  and remove self.get_flattened_size

File: vulcanai2/models/cnn.py (hunk 1 above)
Comment (priyatharsan) [reply vulcanai2/models/cnn.py]: Yeah I thought the same at first. But, since `nn.ModuleList` is more like a list not sure if it will support specifying names similar to `nn.Sequential`, however, it looks like there is a way: https://github.com/pytorch/pytorch/issues/3499

File: vulcanai2/models/dnn.py (hunk 3 above)
Comment (rfratila) [reply vulcanai2/models/dnn.py]: I see. ill give it a shot!

File: vulcanai2/models/basenetwork.py [hunk 4]
```diff
@@ -132,6 +122,16 @@ def __init__(self, name, config, in_dim=None, save_path=None,
             validation_accuracy=[]
         )
 
+        if in_dim is not None:
```
nirtiac: would be better to be `if in_dim:`, no? what if they give an empty list?

rfratila: yup will fix

vulcanai2/models/basenetwork.py
```diff
@@ -140,10 +134,35 @@ def __init__(self, name, config, in_dim=None, save_path=None,
...
+
+        """
+        if self.input_networks is None:
+            self.input_networks = nn.ModuleDict()
+        assert isinstance(in_network, BaseNetwork)
+        assert in_network is not self
+        assert isinstance(self.input_networks, nn.ModuleDict)
+        self.input_networks[in_network.name] = in_network
+        self.in_dim = self._get_in_dim()
+
```
nirtiac: Just a question - will this affect any GPU/CPU stuff?

vulcanai2/models/basenetwork.py
```diff
@@ -168,40 +187,56 @@ def forward(self, inputs, **kwargs):
         if not isinstance(inputs, list):
             inputs = [inputs]
 
-        if self.input_networks is not None:
+        if self.input_networks:
             net_outs = []
-            for net, x in zip(self.input_networks, inputs):
-                net_outs.append(net(x))
+            for in_net, x in zip(self.input_networks.values(), inputs):
```
nirtiac: could you add a comment describing the concept here? took me a while to parse.

vulcanai2/models/basenetwork.py
nirtiac: the input_network function I mean. The fact that you can now add in input networks.

rfratila: that's something we'll need to explore in the GPU PR. this is just implementing it the same way. It may get interesting if your input networks are distributed across different GPU's but as long as the user specifies the right location it shouldn't be that much of a problem. @priyatharsan thoughts?

vulcanai2/models/basenetwork.py
rfratila: sure

PR #78, opened 2018-11-04
GPU/CPU device switch issue fix

### LINKED ISSUE #39: Make it more strict for when the data and models should be offloaded to gnu() and cpu()

priyatharsan:
closes #39

### DIFF
--- a/vulcanai2/models/basenetwork.py
+++ b/vulcanai2/models/basenetwork.py
@@ -9,7 +9,7 @@ import sys
 
 # Vulcan imports
 from .layers import *
-from .utils import network_summary
+from .utils import network_summary, set_device, convert_tensor
 from .metrics import Metrics
 from ..plotters.visualization import display_record
 
@@ -85,16 +85,6 @@ class BaseNetwork(nn.Module):
         super(BaseNetwork, self).__init__()
 
         self._name = name
-        if in_dim is not None:
-            if isinstance(in_dim, int):
-                self.in_dim = tuple([in_dim])
-            else:
-                self.in_dim = in_dim
-        else:
-            if input_networks is None:
-                raise ValueError("BaseNetwork must have either in_dim or \
-                                input_networks")
-
         self._config = config
         self._save_path = save_path
 
@@ -131,6 +121,16 @@ class BaseNetwork(nn.Module):
             validation_error=[],
             validation_accuracy=[]
         )
+
+        if in_dim is not None:
+            if isinstance(in_dim, int):
+                self.in_dim = tuple([in_dim])
+            else:
+                self.in_dim = in_dim
+        else:
+            if input_networks is None:
+                raise ValueError("BaseNetwork must have either in_dim or \
+                                input_networks")
 
         # Creates the Network, and re-writes the self.in_dim
         self._create_network(
@@ -176,31 +176,45 @@ class BaseNetwork(nn.Module):
         else:
             output = torch.cat(inputs, dim=1)
 
-        network_output = self.network(output)
-
-        if self._num_classes:
-            class_output = self.network_tail(network_output)
-            return class_output
-        else:
-            return network_output
+        return self.network(output)
 
     def _get_out_dim(self):
         """
-        Return the shape of the output of network by performing 
-        a single forward pass using built-up data.
+        Return the network output shape.
+
+        Perform a single forward pass using made-up data.
 
         Returns
         -------
         shape : tuple
             The output shape of the network.
+
         """
         if self.network is not None:
             out = self.network(torch.ones([1, *self.in_dim]))
-            if self._num_classes:
-                out = self.network_tail(out)
             return tuple(out.shape[1:])
         else:
             return None
+
+    def _get_in_dim(self):
+        """
+        Return the network input shape.
+
+        Perform a single forward pass through all the input networks
+        and merge together to get input shape of this network.
+
+        Returns
+        -------
+        shape : tuple
+            The input shape of the network.
+
+        """
+        # Create empty input tensors
+        in_tensors = []
+        for d in self.input_networks:
+            in_tensors.append(torch.ones([1, *d.out_dim]))
+        output = self._merge_input_network_outputs(in_tensors)
+        return tuple(output.shape[1:])
 
     @property
     def name(self):
@@ -361,12 +375,6 @@ class BaseNetwork(nn.Module):
             # If freeze is True, set requires_grad to False
             # If freeze is False, set requires_grad to True
             params.requires_grad = not freeze
-        # Freeze prediction layer parameters
-        if 'network_tail' in self._modules:
-            for params in self.network_tail.parameters():
-                # If freeze is True, set requires_grad to False
-                # If freeze is False, set requires_grad to True
-                params.requires_grad = not freeze
         # Recursively toggle freeze on
         if apply_inputs and self.input_networks is not None:
             for network in self.input_networks:
@@ -389,7 +397,8 @@ class BaseNetwork(nn.Module):
         self.criterion = self._init_criterion(self._criter_spec)
 
     def fit(self, train_loader, val_loader, epochs,
-            retain_graph=None, valid_interv=4, plot=False):
+            retain_graph=None, valid_interv=4, plot=False,
+            device=None, non_blocking=False):
         """
         Trains the network on the provided data.
 
@@ -405,6 +414,10 @@ class BaseNetwork(nn.Module):
             Whether retain_graph will be true when .backwards is called.
         valid_interv : int
             Specifies the period of epochs before validation calculation.
+        device : str (eg: "cuda:0" or "cuda:1"...) or torch.device
+            Specifies the desired device to use for data and model.
+        non_blocking : boolean
+            refer: https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.to
 
         Returns
         -------
@@ -425,13 +438,17 @@ class BaseNetwork(nn.Module):
             for epoch in trange(epochs, desc='Epoch: ', ncols=80):
 
                 train_loss, train_acc = self._train_epoch(train_loader,
-                                                          retain_graph)
+                                                          retain_graph,
+                                                          device,
+                                                          non_blocking)
                 if self.lr_scheduler is not None:
                     self.lr_scheduler.step(epoch=epoch)
 
                 valid_loss = valid_acc = np.nan
                 if epoch % valid_interv == 0:
-                    valid_loss, valid_acc = self._validate(val_loader)
+                    valid_loss, valid_acc = self._validate(val_loader,
+                                                           device,
+                                                           non_blocking)
 
                 tqdm.write("\n Epoch {}:\n"
                            "Train Loss: {:.6f} | Test Loss: {:.6f} |"
@@ -461,7 +478,8 @@ class BaseNetwork(nn.Module):
                 "\n\n**********KeyboardInterrupt: "
                 "Training stopped prematurely.**********\n\n")
 
-    def _train_epoch(self, train_loader, retain_graph):
+    def _train_epoch(self, train_loader, retain_graph,
+                     device, non_blocking):
         """
         Trains the network for 1 epoch.
 
@@ -469,6 +487,10 @@ class BaseNetwork(nn.Module):
         ----------
         train_loader : DataLoader
             The DataLoader object containing the dataset to train on.
+        device : str (eg: "cuda:0" or "cuda:1"...) or torch.device
+            Specifies the desired device to use for data and model.
+        non_blocking : boolean
+            refer: https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.to
 
         Returns
         -------
@@ -484,18 +506,15 @@ class BaseNetwork(nn.Module):
 
         for batch_idx, (data, targets) in enumerate(train_loader):
 
-            # for idx, d in enumerate(data):
-            #     data[idx] = Variable(d, requires_grad=True)
-            # targets = Variable(targets)
-
-            if torch.cuda.is_available():
-                for idx, d in enumerate(data):
-                    data[idx] = d.cuda()
-                targets = targets.cuda()
-                self.cuda()
+            if torch.cuda.is_available() and device is not None:
+                self.to(device=device, non_blocking=non_blocking)
+                data = set_device(data, device=device,
+                                  non_blocking=non_blocking)
+                targets = convert_tensor(targets, device=device,
+                                     non_blocking=non_blocking)
 
             # Forward + Backward + Optimize
-            # import pudb; pu.db
+
             predictions = self(data)
 
             train_loss = self.criterion(predictions, targets)
@@ -528,7 +547,7 @@ class BaseNetwork(nn.Module):
 
         return train_loss, train_accuracy
 
-    def _validate(self, val_loader):
+    def _validate(self, val_loader, device, non_blocking):
         """
         Validate the network on the validation data.
 
@@ -536,6 +555,10 @@ class BaseNetwork(nn.Module):
         ----------
         val_loader : DataLoader
             The DataLoader object containing the dataset to evaluate on
+        device : str (eg: "cuda:0" or "cuda:1"...) or torch.device
+            Specifies the desired device to use for data and model.
+        non_blocking : boolean
+            refer: https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.to
 
         Returns
         -------
@@ -551,15 +574,12 @@ class BaseNetwork(nn.Module):
 
         for batch_idx, (data, targets) in enumerate(val_loader):
 
-            # for idx, d in enumerate(data):
-            #     data[idx] = Variable(d, requires_grad=False)
-            # targets = Variable(targets, requires_grad=False)
-
-            if torch.cuda.is_available():
-                for idx, d in enumerate(data):
-                    data[idx] = d.cuda()
-                targets = targets.cuda()
-                self.cuda()
+            if torch.cuda.is_available() and device is not None:
+                self.to(device=device, non_blocking=non_blocking)
+                data = set_device(data, device=device,
+                                  non_blocking=non_blocking)
+                targets = convert_tensor(targets, device=device,
+                                     non_blocking=non_blocking)
 
             predictions = self(data)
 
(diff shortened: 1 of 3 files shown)


vulcanai2/models/basenetwork.py
```diff
@@ -635,9 +656,10 @@ def forward_pass(self, data_loader, convert_to_class=False):
         pred_collector = torch.tensor([])
         for batch_idx, (*data, _) in enumerate(data_loader):
 
-            if torch.cuda.is_available():
-                data = data.cuda()
-                self.cuda()
+            if torch.cuda.is_available() and device is not None:
```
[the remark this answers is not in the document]
priyatharsan: specifying `torch.device("cpu")` or `None` is both same in terms of pytorch's `to` method. They get rendered to `cpu` in the end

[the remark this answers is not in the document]
rfratila: That's fine, but we shouldn't be checking that many times if CUDA is available. if we define the device in only one spot and just refer to that, I think the code is cleaner

vulcanai2/models/utils.py
```diff
@@ -210,3 +210,52 @@ def print_model_structure(network, input_size=None):
...
+
+    """
+    for idx, d in enumerate(data):
+        if isinstance(d, (list, tuple)):
+            data[idx]= set_device(d, device=device,
+                                   non_blocking=non_blocking)
+        else:
+            data[idx] = convert_tensor(d, device=device,
+                                    non_blocking=non_blocking)
+    return data
```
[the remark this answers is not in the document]
priyatharsan: its same as `data = data.cuda()`. I think only for the model you don't have to return.

vulcanai2/models/basenetwork.py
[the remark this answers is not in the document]
priyatharsan: Hmm we could move `if torch.cuda.is_available()` contition to user facing side, but I remember you wanted to reduce the usage of torch methods explicitly in user facing side. That's why I didn't think about that.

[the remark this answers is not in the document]
rfratila: and just always do `.to(device)` without checking anything

vulcanai2/models/utils.py
```diff
@@ -210,3 +210,52 @@ def print_model_structure(network, input_size=None):
...
+    data : list
+        data converted to the relevant device
+
+    """
+    for idx, d in enumerate(data):
+        if isinstance(d, (list, tuple)):
+            data[idx]= set_device(d, device=device,
+                                   non_blocking=non_blocking)
+        else:
+            data[idx] = convert_tensor(d, device=device,
```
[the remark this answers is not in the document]
priyatharsan: I thought the same but we need it for `targets` as well, right? Since targets is not a list at the moment.

vulcanai2/models/utils.py
```diff
@@ -210,3 +210,52 @@ def print_model_structure(network, input_size=None):
         if isinstance(v, odict):
             for k2, v2 in v.items():
                 print('\t {}: {}'.format(k2, v2))
+
+def set_device(data, device=None, non_blocking=False):
```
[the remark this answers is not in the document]
priyatharsan: yeah I will remove that.

vulcanai2/models/basenetwork.py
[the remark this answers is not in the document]
rfratila: no I don't want to push that to user side. I just want to hide it as much as possible. Try specifying `self.device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")` in the __init__ and just refer to `.to(self.device)`

vulcanai2/models/utils.py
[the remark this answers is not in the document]
rfratila: just merge the two functions into one so you can pass both `input_data` and `targets` to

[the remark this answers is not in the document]
rfratila: call it `set_tensor_device`

vulcanai2/models/basenetwork.py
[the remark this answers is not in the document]
priyatharsan: okay. I didn't we agreed on specifying the device in `_init_`, anyway, you want to hard-code the device id?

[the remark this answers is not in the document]
rfratila: It's just a thought but I just want to make it as simple as possible. I don't want to hard code the device id but then the user needs to remember to specify their device ID every time they `.fit` or `.forward_pass`

rfratila:
Im curious what @nirtiac thinks about this^

vulcanai2/models/basenetwork.py
```diff
@@ -140,6 +141,8 @@ def __init__(self, name, config, in_dim=None, save_path=None,
         # Compute self.out_dim of the network
         self.out_dim = self._get_out_dim()
 
+        self.device = torch.device(device if torch.cuda.is_available() else "cpu")
```
rfratila: if cuda is available but the user forgets to specify the device it will still set it to the `cpu`? perhaps we should default `device='cuda:0'` instead @priyatharsan ?

rfratila: if cuda is not available then it will default to cpu anyway

priyatharsan: hmm.. but what if its the case, if the user still would like to stick with `cpu`

rfratila: they can always override `model.device = 'cpu'` after they've created the network. they would explicitly need to specify using the cpu if a cpu is available

nirtiac: Why don't we have the setter for self.device written so that anytime the user changes self.device it calls all the necessary self.to(self.device) etc statements? Or is there something I'm missing?

nirtiac:
Oh interesting, that's a good point. However, I'm not sure I understand why it matters to us if nn.Module doesn't have the device parameter, as the user never accesses nn.Module? They only ever interact with basenetwork, no?

vulcanai2/models/basenetwork.py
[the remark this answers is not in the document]
nirtiac: we could make the device id part of the setter, so if someone wants to get fancy they can get fancy

rfratila:
yeah I like the setter idea that @nirtiac suggested. the user shouldn't care about if the internal nn.Module's have a device component

vulcanai2/models/basenetwork.py
rfratila: I think that can work. @priyatharsan ?

rfratila: also should it extend to all `self.input_networks` as well? similar to how when you call `freeze` you have the option to also freeze incoming networks? @priyatharsan

priyatharsan: I will give it a shot on setter today. That shouldn't be a problem. 

And we don't have to extend to all `self.input_networks`. The network is a nn.ModuleList (currently in this branch) so it supports .to(device) so setting it will, by default, switches the device internally to its own input_networks. I need to check how it reacts with nn.ModuleDict. However, those input_networks will be, individually, in the default device (cpu). Eg: setting `multi_input_conv_3D` to gpu will set itself and its input_networks to gpu but not set `conv_3D` to gpu unless we specify them manually. Do you get me?

rfratila: im pretty sure ModuleList and ModuleDict work the same way in terms of the `.to` functionality. Just to clarify your example, if `conv_3D` is an input to the `multi_input_conv_3D`, and you set the device of the latter to gpu, then it will also change the former?

priyatharsan: no it won't explicitly but internally yes. To be more clear, if we set `self.to(device`) it sets `self.network` and `self.input_networks = nn.ModuleList(list of input_networks that are nn.Modules)` `.to(device)` but not individually on each individual `input_networks`. Does that makes sense?

priyatharsan: we will discuss this after the meeting today. If its confusing 😅

rfratila: I see. But I'm assuming that in addition to just the ModuleList the embedded networks will also be run on the specified device? maybe we should explicitly set it instead of the implicit assumption?

priyatharsan: >I'm assuming that in addition to just the ModuleList the embedded networks will also be run on the specified device?
yep! they run on the specified device assigned to nn.ModuleList.

>maybe we should explicitly set it instead of the implicit assumption?
hmm... we can do that by looping over individual input_networks just like we did before when used `list` object instead of `nn.ModuleList`, remember `_to_cuda` func?. That's an additional step, but is that necessary?

rfratila: we can confirm that simply setting the `ModuleList` to a device that the internal networks also get assigned to that device during run time. that would clear up my concerns

priyatharsan: Yep it works. The current version in this branch trains in my gpu without needing to individually switch the device for the input_networks. Is it possible you can set up the ml server to cuda ready today?

vulcanai2/models/cnn.py
```diff
@@ -124,12 +124,13 @@ def __init__(self, name, config, in_dim=None, save_path=None,
                  activation=nn.ReLU(), pred_activation=None,
                  optim_spec={'name': 'Adam', 'lr': 0.001},
                  lr_scheduler=None, early_stopping=None,
-                 criter_spec=nn.CrossEntropyLoss()):
+                 criter_spec=nn.CrossEntropyLoss(),
+                 device="cpu"):
```
rfratila: let's make sure the defaults of `cuda:0` are consistent.

vulcanai2/models/dnn.py
```diff
@@ -114,12 +114,13 @@ def __init__(self, name, config, in_dim=None, save_path=None,
                  activation=nn.ReLU(), pred_activation=None,
                  optim_spec={'name': 'Adam', 'lr': 0.001},
                  lr_scheduler=None, early_stopping=None,
-                 criter_spec=nn.CrossEntropyLoss()):
+                 criter_spec=nn.CrossEntropyLoss(),
+                 device="cpu"):
```
rfratila: same: let's make sure the defaults of `cuda:0` are consistent.

vulcanai2/models/basenetwork.py
```diff
@@ -612,7 +646,8 @@ def cross_validate(self, data_loader, k, epochs,
             figure_path=figure_path)  # TODO: deal with repeated default parameters
 
     # TODO: Instead of self.cpu(), use is_cuda to know if you can use gpu
-    def forward_pass(self, data_loader, convert_to_class=False):
+    def forward_pass(self, data_loader, convert_to_class=False,
+                     device=None, non_blocking=False):
```
rfratila: I think you may have forgotten to remove the `non_blocking` parameter

vulcanai2/models/basenetwork.py
```diff
@@ -551,15 +592,8 @@ def _validate(self, val_loader):
 
         for batch_idx, (data, targets) in enumerate(val_loader):
 
-            # for idx, d in enumerate(data):
-            #     data[idx] = Variable(d, requires_grad=False)
-            # targets = Variable(targets, requires_grad=False)
-
-            if torch.cuda.is_available():
```
rfratila: @priyatharsan don't you need to also move the optimizer to the specified device? or is that implicitly done.

vulcanai2/models/basenetwork.py
priyatharsan: yeah I forgot to remove that

vulcanai2/models/cnn.py
priyatharsan: oops yeah its removed now

vulcanai2/models/dnn.py
priyatharsan: 👍

vulcanai2/models/basenetwork.py
priyatharsan: Yeah it is done implicitly, I just checked it. You can get the device by loop through the optimizer's state_dict, something like: 
```
for state in optimizer.state.values():
    for k, v in state.items():
        if torch.is_tensor(v):
            #state[k] = v.is_cuda
            state[k] = v.device
```

priyatharsan:
I need to check with the save and load scenarios. Will do tonight.

vulcanai2/models/basenetwork.py
rfratila: there might be issues when switching devices in between fit sessions from what I read online. try to exhaustively test all the variations and create unit tests for what you can

priyatharsan: wait! we are not switching device between fit sessions, do we? also, can you share the resource you are referring to?

rfratila: What Im referring to is being able to train on gpu and then train on cpu if needed, and vice versa. This issue explains it clearly: https://github.com/pytorch/pytorch/issues/2830

vulcanai2/models/basenetwork.py
```diff
@@ -144,8 +144,7 @@ def __init__(self, name, config, in_dim=None, save_path=None,
         # Compute self.out_dim of the network
         self.out_dim = self._get_out_dim()
 
-        self._device = torch.device(device if torch.cuda.is_available() else "cpu")
-        self.to(device=self._device)
```
priyatharsan: why remove this? when you set device don't you have to cast the network to the device? It's causing issues now

rfratila: because we had duplicated the setter function in the init. I just simplified it by calling the device setter which runs those lines of code. Would you be able to provide the exact error/issue that you're running into?

priyatharsan: It used to work this way:
when we set `multi_input_conv_3D.device = "cpu"`
it will set to via  `multi_input_conv_3D.input_networks[#].device = "cpu"`, remember I showed you the other day?
Now, it doesn't seem to be working. I have to loop into individual input_networks in this case. Just came across this when testing with a loaded model.

rfratila: you can try to add in the device init code back but honestly its the same code as the device.setter. however, I don't think we should set all input networks to the same device as the current network since users may have already specified them to a specific cpu and don't want it to be moved again. we can add an option to the setter to also change the input networks devices

priyatharsan: I think `torch.distributed` would do a better job if we are considering multiple device scenario.
Also, if you'd like to have the input_networks in the default device unless specified, how do you plan to assign the data's device based on the network's device? Only option I can think of is looping through each input_network and check for the device and assign that device to the incoming data, whichcan done within forward function but I'm not sure if that is recommended.

rfratila: torch.distributed is for training across various machines. we are still on the same machine with different gpus. I just delayed the data device casting until the forward function which seems to work pretty well especially if users change network gpu devices in real-time

vulcanai2/models/basenetwork.py
rfratila: I think the best way to handle it is just to rebuild the optimizer instead of distributing the param_groups to the devices of the respective networks?

vulcanai2/models/basenetwork.py
priyatharsan: I mean dataparallel. It looks like the suggested way is to assign the device to data in forward function: https://pytorch.org/tutorials/beginner/former_torchies/parallelism_tutorial.html

vulcanai2/models/basenetwork.py
rfratila: but it's not ideal since you forget the state_dict of the optimizer which isn't ideal for continuing to train

priyatharsan: That is the recommended way so far. But we could try experimenting with the optimi.param.device

priyatharsan:
Now the model can run in multiple devices (even after save and load with different device assigned)

rfratila:
When trying to train the `loaded_model` completely on the CPU I get this error @priyatharsan: 
screen shot 2018-11-10 at 3 35 01 pm

priyatharsan:
Hmm....I think this is related to recreating the optimizer stuff (after loading). I will take a look at it.

rfratila:
Also please add some gpu specific unit tests that only run if a gnu is present

rfratila:
When you removed the check if `torch.cuda.is_available()` in the device.setter() it broke nearly all of the test cases on cpu machines. Please revert back to only set the device to gnu if cuda is available @priyatharsan

rfratila:
Also be careful with `self.network.to` calls when it is None.

priyatharsan:
Need to add unit tests yet...

vulcanai2/models/basenetwork.py
```diff
@@ -235,11 +232,26 @@ def device(self):
...
+            The device to transfer network to.
+
         """
-        self.network.to(device=device)
-    
+        device = torch.device(device if torch.cuda.is_available() else 'cpu')
+        if self.network:
+            self.network.to(device=device)
+        if self.optim:
+            for state in self.optim.state.values():
```
rfratila: If we have an input network that is on a separate device, and we set the optima parameters to the current device, won't that return an error?

rfratila: Ok I can confirm that this crashes if one of the inputs is on another device. It only works if they are all on the same device

rfratila: Also I think we should handle nn.DataParallel for the users if possible

vulcanai2/tests/models/test_device.py
```diff
@@ -139,10 +140,31 @@ def test_master_net_device_set_to_cuda(self, multi_net):
...
-    # @pytest.mark.xfail
-    # def test_fail_mixed_devices(self, multi_net):
-    # TODO: write script with mixed devices so that the 
-    # network fails when fitting
+    def test_fail_mixed_devices(self, multi_net):
+        master_device_setter(multi_net, 'cuda:0')
+        multi_net.input_networks['dense_net'].input_networks['conv2D_net'].device = "cpu"
+
+        data = MultiDataset([
+            (TensorDataset(torch.ones([10, *multi_net.input_networks['conv3D_net'].in_dim])), True, False),
```
rfratila: @priyatharsan any way to shorten the call to input networks? perhaps having a ref link to them directly?

priyatharsan: yeah that could work but that will eventually increases the parameters to the function, is that okay?

rfratila: that's fine. the functions aren't meant to be used by anyone but the testing environment. Plus you can also do assertions that the model references are the same. e.g. `assert conv_2D_net == multi_input.input_networks['conv_2D_net']`

PR #91, opened 2018-11-12
run_test modularization

### LINKED ISSUE #68: re-write run test into separate functions

nirtiac:
Here's a first pass for review. A few points:

1. Metrics doesn't really need to be a class
2. Is the calculation of average correct? shouldn't we have TN in there as well??
3. Still need to do error handling, improve some documentation. 
4. I'm not a huge fan of this, I feel all the parameter passing increasing risk of programmer/maintainer error. 
5. I probably messed up some types in my documentation, will fix. 

Fixes #68

### DIFF
--- a/vulcanai2/models/metrics.py
+++ b/vulcanai2/models/metrics.py
@@ -91,7 +91,7 @@ class Metrics(object):
 
         """
         if metric == 'accuracy':
-            # TODO: Use get_class
+            # TODO: Use extract_class_labels
             max_index = predictions.max(dim=1)[1]
             correct = (max_index == targets).sum() # TODO: this doesn't seem correct
             accuracy = int(correct.data) / len(targets)
@@ -102,7 +102,7 @@ class Metrics(object):
 
     # TODO: class # should correspond with self.num_class
     # noinspection PyMethodMayBeStatic
-    def get_class(self, in_matrix):
+    def extract_class_labels(self, in_matrix):
         """
         Reformat truth matrix to be the classes in a 1D array.
 
@@ -152,13 +152,13 @@ class Metrics(object):
             raise ValueError('There\'s no classification layer')
 
         # getting just the y values out of the dataset
-        test_y = np.array([v[1] for v in data_loader.dataset])
+        test_y = np.array([v[1] for v in data_loader.dataset]) #TODO: store in tensor for continuity?
 
         raw_prediction = network.forward_pass(
             data_loader=data_loader,
             convert_to_class=False)
 
-        class_prediction = self.get_class(raw_prediction)
+        class_prediction = self.extract_class_labels(raw_prediction)
 
         confusion_matrix = get_confusion_matrix(
             predictions=class_prediction,
--- a/vulcanai2/models/metrics.py
+++ b/vulcanai2/models/metrics.py
@@ -39,7 +39,7 @@ class Metrics(object):
         ----------
         targets: numpy.ndarray of integers
             the target values
-        predictions: numpy.ndarray of integers
+        predictions: numpy.ndarray of integers or numpy.ndarray of floats
             the predicted values
         metrics: list of strings
             the strings definition the "get_"... functions to call
@@ -54,9 +54,9 @@ class Metrics(object):
             extract_class_labels
             False if raw_predictions are used
 
-        Returns dict
-            Metric name : values returned by the metric functions
+        Returns
         -------
+        Dict of metric : values returned by the metric functions
 
         """
 
@@ -74,15 +74,15 @@ class Metrics(object):
             method_name = "get_" + metric
             method = getattr(Metrics, method_name)
             try:
-                if method_name in double_parametered_functions:
-                    res = method(targets, predictions)
+                if method_name in set(double_parametered_functions):
+                    metric_results = method(targets, predictions)
                 else:
-                    res = method(targets, predictions, average)
+                    metric_results = method(targets, predictions, average)
 
-            except AttributeError:
+            except (AttributeError,  TypeError):
                 logger.warning("Metric {} does not exist".format(metric))
 
-            results_dict[metric] = res
+            results_dict[metric] = metric_results
 
         return results_dict
 
@@ -116,9 +116,9 @@ class Metrics(object):
     def get_confusion_matrix_values(targets, predictions):
         """
         Will calculate the tp, tn, fp, fn values given targets and predictions
+
         Parameters
         ----------
-
         targets: numpy.ndarray of integers
             the target values
 
--- a/vulcanai2/models/metrics.py
+++ b/vulcanai2/models/metrics.py
@@ -348,7 +348,7 @@ class Metrics(object):
         Calculate the accuracy
         Parameters
         ----------
-       predictions: numpy.ndarray of integers
+        predictions: numpy.ndarray of integers
         the predicted values
 
         targets: numpy.ndarray of integers
@@ -361,7 +361,7 @@ class Metrics(object):
         The accuracy
 
         """
-        accuracy = skl_metrics.accuracy_score(predictions, targets)
+        accuracy = skl_metrics.accuracy_score(targets, predictions)
 
         return accuracy
 
--- a/vulcanai2/models/metrics.py
+++ b/vulcanai2/models/metrics.py
@@ -126,6 +126,186 @@ class Metrics(object):
         elif in_matrix.shape[1] == 1:
             return np.around(in_matrix)
 
+    # TODO: check types
+    def get_confusion_matrix_values(self, confusion_matrix):
+        """
+        Will calculate the tp, tn, fp, fn values given a confusion matrix
+        Parameters
+        ----------
+        confusion_matrix: sklearn.metrics.confusion_matrix
+        The confusion matrix
+
+        Returns
+        -------
+        tp, tn, fp, fn:  np.float32
+        The values calculated
+        """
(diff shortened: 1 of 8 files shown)


rfratila:
Some initial comments @nirtiac:

1) Yeah the way we are currently using doesn't really make sense for it to be a class. perhaps we should store the confusion matrix in the class so we don't need to pass it around? But then the question of how do you know when the class confusion matrix becomes stale and needs to be recalculated.

2) Accuracy should be (TP + TN) / (TP+TN+FP+FN) so yeah we need to include the TN

4) It's not ideal to have different interfaces with the different functions. I think I would rather pass in the confusion matrix to all of them and they should be able to calculate it. But then that means we duplicate some calculations. However that does provide a lot more flexibility in the order that you can call the functions. Separating them completely would be better? Im curious what your thoughts are on this.

nirtiac:
@rfratila Yeah, I totally agree about the interfaces thing and the order of computation issue, I was very torn while writing it. But I realize now that it's rare we're going to have so much data that these are going to be huge costly operations, plus its all numpy, so maybe it won't matter that we duplicate. What's the biggest amt of data we'll have for these as a ballpark? I could run some quick speed tests.

rfratila:
The most number of classes that I can realistically think of would be on the order of 10000s (total number of medications and treatments possible) but even then that's not a lot if we reach 100000s of classes. Now that we have stand-alone functions, do you think that you can integrate them all with `get_metrics()` so eventually users would be able to specify the metrics to track during training?

nirtiac:
Ok, so I'll re-write to have consistent interface, and yes I can do that, I'll push a prototype of the get_metrics integration late today or tomorrow.

rfratila:
Awesome thanks!

vulcanai2/models/metrics.py
```diff
@@ -126,6 +126,186 @@ def extract_class_labels(self, in_matrix):
         elif in_matrix.shape[1] == 1:
             return np.around(in_matrix)
 
+    # TODO: check types
+    def get_confusion_matrix_values(self, confusion_matrix):
```
nirtiac: @rfratila what do you think, should this be one function or four?

rfratila: like `get_tp()`, `get_fp(`), etc? you do need some to calculate others so I think it's best to keep these together?

nirtiac:
@rfratila done, need to test. Notice the auc function, however, I had to change the interface. Need to look into de-constructing the raw values?? I could also get num_classes if needed. Not sure, what do you think?

vulcanai2/models/metrics.py
```diff
@@ -126,185 +86,327 @@ def extract_class_labels(self, in_matrix):
...
-            [np.sum(confusion_matrix)] *
-            confusion_matrix.shape[0]) -
-            confusion_matrix.sum(axis=0) -
-            confusion_matrix.sum(axis=1) + tp).astype('float32')
-        # sum each column and remove diagonal
-        fp = (confusion_matrix.sum(axis=0) - tp).astype('float32')
-        # sum each row and remove diagonal
-        fn = (confusion_matrix.sum(axis=1) - tp).astype('float32')
+
+        tn, fp, fn, tp = confusion_matrix(targets, predictions).ravel()
```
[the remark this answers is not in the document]
rfratila: Yeah I played around with it briefly and it doesn't seem to generalize well to multi-class. Ravel just linearizes the conf matrix which only makes sense in the binary case. Is there another way to use sklearn to get these 4 values without reverting back to what we had before?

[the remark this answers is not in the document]
nirtiac: oooh good catch ok, I will look into this now.

vulcanai2/models/metrics.py
```diff
@@ -126,185 +86,327 @@ def extract_class_labels(self, in_matrix):
...
         Returns
         -------
-        specificity: array of np.float32
+        specificity: np.float32 or array of np.float32
         The specificity
         """
 
-        specificity = np.nan_to_num(tn / (tn + fp))
+        Metrics._check_average_parameters(targets, predictions, average)
+        specificity = precision_score(targets, predictions, average)
```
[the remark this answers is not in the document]
nirtiac: oops 😶

vulcanai2/models/metrics.py
```diff
@@ -126,185 +86,327 @@ def extract_class_labels(self, in_matrix):
...
         Returns
         -------
-        f1: array of np.float32
-        the f1 score
+        f1: np.float32 or array of np.float32
+        The f1 score
         """
 
-        f1 = np.nan_to_num(2 * (ppv * sensitivity) / (ppv + sensitivity))
+        Metrics._check_average_parameters(targets, predictions, average)
```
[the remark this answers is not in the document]
nirtiac: oh nice

vulcanai2/models/metrics.py
```diff
@@ -126,185 +86,327 @@ def extract_class_labels(self, in_matrix):
...
+        f1 = f1_score(targets, predictions, average)
         return f1
 
-    def get_f1_macro(self, ppv, sensitivity):
+    @staticmethod
+    def get_auc(targets, raw_predictions, num_classes, average=None):
```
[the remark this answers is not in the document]
rfratila: say if `raw_predictions = [0.1, 0.8]` for 2 samples in the binary case or `raw_predictions = [[0.2, 0.8], [0.9, 0.1]]`for 2 samples in the multi-class case would the num_classes be the same? or 1 and 2, respectively for the example?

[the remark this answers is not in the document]
nirtiac: oh damn, I see. I think we should keep num_classes in then; we're already breaking the interface with the raw_predictions, it's a completely different matrix.

vulcanai2/models/metrics.py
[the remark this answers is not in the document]
nirtiac: Update: have not found a better solution, so have just reverted to what you had.

nirtiac:
@rfratila ready for re-review. more tests to come, but the class itself is all there.

vulcanai2/models/metrics.py
```diff
@@ -28,81 +28,66 @@ class Metrics(object):
...
-            Network output of shape [batch, num_classes].
-        targets : torch.LongTensor
-            The truth values of shape [batch].
-        metric : str
-            The metric to calculate and return.
-
-        Returns
+        targets: numpy.ndarray of integers
+            the target values
+        predictions: numpy.ndarray of integers
```
rfratila: These can be raw_predictions as well so it can be either integers or floats.

vulcanai2/models/metrics.py
```diff
@@ -28,81 +28,66 @@ class Metrics(object):
...
+            This parameter is required for multiclass/multilabel targets.
+            If None, the scores for each class are returned.
+            Otherwise, this determines the type of averaging performed on the
+            data.See scikit learn
+        class_converted: binary. default False
+            True if raw_predictions have already been converted using
+            extract_class_labels
+            False if raw_predictions are used
+
+        Returns dict
```
rfratila: the underline `-------` should be directly underneath "Returns"

vulcanai2/models/metrics.py
```diff
@@ -28,81 +28,66 @@ class Metrics(object):
...
+        if isinstance(metrics, str):
+            metrics = [metrics]
+
+        results_dict = {}
+
+        for metric in metrics:
+            method_name = "get_" + metric
+            method = getattr(Metrics, method_name)
+            try:
+                if method_name in double_parametered_functions:
```
rfratila: ```if method_name in set(double_parametered_functions):``` may be faster

vulcanai2/models/metrics.py
```diff
@@ -28,81 +28,66 @@ class Metrics(object):
...
+                    res = method(targets, predictions)
```
rfratila: can we use `metric_results` as the name instead of `res`?

vulcanai2/models/metrics.py
```diff
@@ -28,81 +28,66 @@ class Metrics(object):
...
+                else:
+                    res = method(targets, predictions, average)
+
+            except AttributeError:
```
rfratila: Should we also try to catch for some sort of not callable error?

vulcanai2/models/metrics.py
```diff
@@ -126,6 +112,345 @@ def get_class(self, in_matrix):
         elif in_matrix.shape[1] == 1:
             return np.around(in_matrix)
 
+    @staticmethod
+    def get_confusion_matrix_values(targets, predictions):
+        """
+        Will calculate the tp, tn, fp, fn values given targets and predictions
+        Parameters
```
rfratila: There should be a free line above `Parameters`, not beneath it.

vulcanai2/models/metrics.py
nirtiac: add in TypeError and description

vulcanai2/models/metrics.py
```diff
@@ -74,15 +74,15 @@ def get_score(self, targets, predictions, metrics='accuracy', average=None,
...
-                    res = method(targets, predictions)
+                if method_name in set(double_parametered_functions):
+                    metric_results = method(targets, predictions)
                 else:
-                    res = method(targets, predictions, average)
+                    metric_results = method(targets, predictions, average)
 
-            except AttributeError:
+            except (AttributeError,  TypeError):
                 logger.warning("Metric {} does not exist".format(metric))
```
rfratila: Could you include the `method_name` in the error message?

PR #97, opened 2018-11-26
Update README.md

nirtiac:
See here a skeleton prototype for the readme. I wanted to keep it short, so we don't overwhelm potential users, and so we can identify now where we would like to add more text for emphasis. 

I added installation and contribution lines even though we don't have that set up yet. We can discuss those now and then remove them until they become relevant, if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,27 @@
 # Vulcan
-Deep Learning Framework
+Vulcan is Aifred Health's framework for rapid deep learning model prototyping and analysis.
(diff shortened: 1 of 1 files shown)


nirtiac:
@mrkrause any comments?

mrkrause:
Looks good to me. The only thing that's slightly weird is that we have links for only one of the "features". I'll try to scare some more up this afternoon.

PR #101, opened 2018-11-27
Dataset improvements and tests

nirtiac:
Not yet ready for full review. Need to first resolve interface with the new split method and add related tests. Also need to investigate and change the random seed


vulcanai2/datasets/tabulardataset.py
```diff
@@ -15,54 +15,87 @@
...
+            Key(s) that specifies which columns to use to uniquely stitch
+            dataset (default None)
+        index_list: list of strings
+            List of columns to make the index of the dataframe
+        na_values: The values to convert to NaN when reading from csv
+        kwargs: keyword parameter, value is either path or dataframe
+            Where key: dataset name and value is either a path to a file
+            or a dataframe.
+        """
+        dataset_dict = kwargs
```
rfratila: if it's the dataset_dict, shouldn't we call the function parameter that instead of `kwargs`?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -15,54 +15,87 @@
...
+        for dataset in dataset_dict:
+            v = dataset_dict[dataset_dict]
```
rfratila: woah what is happening here? is it a typo which should be `dataset_dict[dataset]`?

rfratila: also let's use a more informative variable name than just `v`

vulcanai2/datasets/tabulardataset.py
```diff
@@ -73,35 +106,60 @@ def __getitem__(self, idx):
...
-    #         if isinstance(index_list, list):
-    #             assert len(data) == len(index_list)
-    #         self.df = utils.stitch_datasets(dfs, join_column)
-    #     elif isinstance(data, pd.DataFrame):
-    #         self.df = utils.stitch_datasets([self.df, data], join_column, index_list)
-    #     else:
-    #         new_df = pd.read_csv(data, na_values=na_values)
-    #         self.df = utils.stitch_datasets([self.df, new_df], join_column, index_list)
-    #     logger.info(f"Datasets successfully merged")
+    def merge_data(self, merge_on_columns=None,
```
rfratila: should we call it `merge_dataframe`?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -73,35 +106,60 @@ def __getitem__(self, idx):
...
+            List of columns to make the index of the dataframe
+        na_values: The values to convert to NaN when reading from csv
+        kwargs: keyword parameter, value is either path or dataframe
+            Where key: dataset name and value is either a path to a file
+            or a dataframe.
+        """
+
+        dataset_dict = kwargs
+
+        if "original" in dataset_dict:
```
rfratila: How do you know that 'original` will be in the dict?

vulcanai2/datasets/tabulardataset.py
nirtiac: ... that is an excellent point. I kinda just, defaulted to kwargs

vulcanai2/datasets/tabulardataset.py
nirtiac: woah, yeah, something happened there. Sounds good re variable name

vulcanai2/datasets/tabulardataset.py
nirtiac: ok!

vulcanai2/datasets/tabulardataset.py
nirtiac: I'm checking to make sure it won't be because I want to name the original dataset "original". Not great, but Josephs scheme requires a name. Other suggestions appreciated!

vulcanai2/datasets/tabulardataset.py
rfratila: kwargs is normally left for other case-by-case parameters that would need to be passed

vulcanai2/datasets/tabulardataset.py
rfratila: perhaps we can simply name them with a zero-indexed int in ascending order if no name is provided?

nirtiac: Well:
a.) you can't not provide a name when merging as we're using keyword arguments
b.) you can't start a pandas column with an integer
c.) we don't have any way of knowing what was used before, which makes me a bit wary of using a, b, c as those are frequently used in variable names (which also negates using "original" if multiple merges are performed)

rfratila: @joemehltretter what are your thoughts on this. Is naming it "original" the best course of action?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -241,37 +259,51 @@ def create_label_encoding(self, column_name, ordered_values):
...
+        ----------
+        column_name: String
+            The name of the column you want to one-hot encode
+        prefix_sep String default("@")
+            The prefix used when creating a one-hot encoding
         """
-        #TODO: ensure dummy_na =False is what you want
+
+        # TODO: ensure dummy_na =False is what you want
         if column_name in list(self.df):
```
rfratila: if you're providing a column name to encode, why loop through?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -241,37 +259,51 @@ def create_label_encoding(self, column_name, ordered_values):
...
-            self.df = pd.get_dummies(self.df, dummy_na=False, columns=[column_name], prefix_sep=prefix_sep)
+            self.df = pd.get_dummies(self.df, columns=[column_name],
+                                     prefix_sep=prefix_sep)
             logger.info(f"Successfully encoded {column_name}")
 
         else:
             logger.info(f"Col {column_name} does not exist")
 
     def reverse_create_all_one_hot_encodings(self, prefix_sep="@"):
```
rfratila: perhaps just `reverse_one_hot_encoding` b/c of similar signature name to the one that encodes them. perhaps you can also pass in a column name to reverse? and default it to do all the ones it finds?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -282,35 +314,56 @@ def reverse_create_all_one_hot_encodings(self, prefix_sep="@"):
...
         """
+
         if threshold >= 1 or threshold <= 0:
-            raise ValueError("Threshold needs to be a proportion between 0 and 1")
+            raise ValueError(
+                "Threshold needs to be a proportion between 0 and 1")
         num_threshold = (threshold * len(self))
-        tmp = self.df.dropna(thresh=num_threshold, axis=1)  # thresh is "Require that many non-NA values."
+        # thresh is "Require that many non-NA values."
+        tmp = self.df.dropna(thresh=num_threshold, axis=1)
```
rfratila: are you sure `dropna` isn't done inline?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -282,35 +314,56 @@ def reverse_create_all_one_hot_encodings(self, prefix_sep="@"):
...
         return cols
 
     def identify_unique(self, threshold):
         """
         Returns columns that have at least threshold number of values
-        :param threshold: All columns having at least threshold number of unique values.
+
+        Parameters
+        ----------
+        threshold: The minimum number of values needed
```
rfratila: can you define the values that threshold should lie between?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -319,46 +372,74 @@ def identify_unique(self, threshold):
...
-        :param threshold: Proportion needed to define unbalanced, between 0 and 1
-        :param non_numeric: Whether non-numeric columns are also considered.
-        :return: The column list
+        This returns columns that are highly unbalanced.
+        Those that have a disproportionate amount of one value.
+
+        Parameters
+        ----------
+        threshold: Float
+            Proportion needed to define unbalanced, between 0 and 1
```
rfratila: does 0 mean very unbalanced or very balanced? let's specify

vulcanai2/datasets/tabulardataset.py
```diff
@@ -319,46 +372,74 @@ def identify_unique(self, threshold):
...
+        -------
+        column_list: List
+            The list of column names
         """
         if non_numeric:
             columns = list(self.df.columns)
         else:
             columns = list(self.df.select_dtypes(include=np.number))
         column_list = []
         for col in columns:
```
rfratila: is there a way to do this without looping through the columns? perhaps pandas has some useful utilities?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -319,46 +372,74 @@ def identify_unique(self, threshold):
...
     def identify_highly_correlated(self, threshold):
         """
         Identify columns that are highly correlated with one-another.
-        :param threshold: Between 0 and 1. Minimum amount of correlation necessary to be identified.
-        :return: None
+
+        Parameters
+        ----------
+        threshold: Between 0 and 1.
+        Minimum amount of correlation necessary to be identified.
```
rfratila: let's specify what 0 and 1 signify

vulcanai2/datasets/tabulardataset.py
```diff
@@ -319,46 +372,74 @@ def identify_unique(self, threshold):
...
+        -------
+        column list: List of tuples
+            The correlation values above threshold
         """
         column_list = set()
-        featuresCorrelation = self.df.corr().abs()
-        corr_pairs = featuresCorrelation.unstack()
+        features_correlation = self.df.corr().abs()
+        corr_pairs = features_correlation.unstack()
         for index, val in corr_pairs.items():
```
rfratila: any pandas functionality that can simplify this logic? something similar to `numpy.where`?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -372,36 +453,50 @@ def identify_low_variance(self, threshold):
...
-    # future improvements could come from https://github.com/pytorch/text/blob/master/torchtext/data/dataset.py
+    # future improvements could come from
+    # https://github.com/pytorch/text/blob/master/torchtext/data/dataset.py
     # noinspection PyUnusedLocal
     def split(self, split_ratio=0.7, stratified=False, strata_field='label',
               random_state=None):
-        """Create train-test(-valid?) splits from the instance's examples.
-        Function signature borrowed from torchtext in an effort to maintain consistency
+        """
+        Create train-test(-valid?) splits from the instance's examples.
```
rfratila: let's write it as "validation" instead of the shortened version in the docs

vulcanai2/datasets/tabulardataset.py
```diff
@@ -372,36 +453,50 @@ def identify_low_variance(self, threshold):
...
+            whether the sampling should be stratified.
                 Default is False.
-            strata_field (str): name of the examples Field stratified over.
-                Default is 'label' for the conventional label field.
-            random_state (int): the random seed used for shuffling.
-        Returns:
-            Tuple[Dataset]: Datasets for train, validation, and
-                test splits in that order, if the splits are provided.
-        """
+        strata_field: String
```
rfratila: should we call it strat**i**_field instead of strat**a**_field

vulcanai2/datasets/tabulardataset.py
joemehltretter: @rfratila default is false for dropna

vulcanai2/datasets/tabulardataset.py
joemehltretter: We might be able to use apply functions, but the readability might not be as nice. For example, `self.df.apply(pd.value_counts)`. However, I still believe looping to be the best option since we need to a couple operations on the columns.

vulcanai2/datasets/tabulardataset.py
rfratila: ok as long as self.df isn't changed this is fine.

vulcanai2/datasets/tabulardataset.py
rfratila: could you time it? if it is significantly slower than I would rather beef up the documentation to understand what is happening behind the scenes

vulcanai2/datasets/tabulardataset.py
nirtiac: Hmm ok, good point x3, will do.

vulcanai2/datasets/tabulardataset.py
rfratila: Since you only care about the correlations that are greater than the threshold. you can use the .where to simply extract those instead of setting the others to 0 and then loop through again. does that make sense? you could probably reduct this to only one line

joemehltretter: @rfratila I don't know if that's possible for dataframe, because it would require removing entire rows and columns. For example, in the attached image in order to remove the zero from that row+column combo we'd have to either remove the entire row or the entire column. We don't want to do that because you can see that some highly correlated values would be lost if we removed. Does that make sense?
corr_example_vulc2

vulcanai2/datasets/tabulardataset.py
joemehltretter: I tested and can confirm it's not done inline

vulcanai2/datasets/tabulardataset.py
joemehltretter: @rfratila So it looks like looping is faster than using apply method. However, using apply method is more concise. Here is the apply method:

```
col_max_val = self.df.apply(pd.value_counts).max()
ls_cols = col_max_val.where(col_max_val >= max_num_vals).dropna().index.values.tolist()
```

vulcanai2/datasets/tabulardataset.py
rfratila: yeah I think that sounds good. If @nirtiac agrees, I think it would merit to prototype this and see what it looks like in practice

vulcanai2/datasets/tabulardataset.py
rfratila: interesting how looping is faster. in that case, no need to use the apply method.

vulcanai2/datasets/tabulardataset.py
rfratila: could you create a temp data frame that collects the correct rows instead of having to remove them from the original if that's not feasible? I'd just like to loop at most once to do this

joemehltretter: @rfratila in the current implementation we are only looping once

vulcanai2/datasets/tabulardataset.py
nirtiac: I like it

vulcanai2/datasets/tabulardataset.py
rfratila: ok great!

vulcanai2/datasets/tabulardataset.py
nirtiac: Good point, I'll let it fail through pandas

vulcanai2/datasets/tabulardataset.py
```diff
@@ -290,37 +289,52 @@ def create_one_hot_encoding(self, column_name, prefix_sep="@"):
...
+    # if a use case presents itself column_name could easily become a list
+    def reverse_create_one_hot_encoding(self, column_name=None,
+                                        reverse_all=False, prefix_sep="@"):
         """
         Undo the creation of one-hot encodings, if prefix_sep was used
-        to create one-hot encodings and nowhere else.
-
+        to create one-hot encodings and nowhere else. If a column_name is
+        provided, only that column will be reverse-encoded, even if reverse_all
+        is set to true.
```
nirtiac: @rfratila let me know if you think it's overkill to have both reverse_all and column_name set to a negative value, so that the user has to provide one or the other

nirtiac: tis a wee bit weird

rfratila: yeah let's just have `column_name` and if it's none it will reverse all, otherwise just the ones specified

nirtiac: cool

joemehltretter:
@rfratila I talked with @nirtiac and we thought it would be best to use `kwargs` for `init`, `merge_dataframe`, and `split_datasets`. I just pushed an updates that reflects that. So, rather than sending a dictionary of dataframes, we are sending keyword arguments. Our talk clarified where I was going wrong, and was able to get it functioning the way it was meant to be. Please let me know what you think. If you approve, then I will update the tests.

vulcanai2/datasets/utils.py
```diff
@@ -154,41 +154,39 @@ def stitch_datasets(df_dict, df_main=None, merge_on_columns=None,
         (default None)
     index_list: list of strings
         columns to establish as index for final stitched dataset (default None)
-
+    kwargs : keyword parameter, value is dataframe
```
rfratila: could you update this to be `dataset_dict` please?

nirtiac:
@joemehltretter @rfratila I believe literally all the comments Robert just made could be resolved by reverting to the commit I made just before Joseph's most recent changes :P

joemehltretter:
> 
> 
> @joemehltretter @rfratila I believe literally all the comments Robert just made could be resolved by reverting to the commit I made just before Joseph's most recent changes :P

Yeah I think so :D minus the stitch dataset and some other minor updates

vulcanai2/datasets/utils.py
joemehltretter: @rfratila Updated

vulcanai2/datasets/tabulardataset.py
```diff
@@ -122,9 +137,8 @@ def merge_dataframe(self, merge_on_columns=None,
...
             List of columns to make the index of the dataframe
         na_values: The values to convert to NaN when reading from csv
-        kwargs: keyword parameter, value is either path or dataframe
-            Where key: dataset name and value is either a path to a file
-            or a dataframe. This parameter should only contain new dataframes.
+        kwargs: keyword parameter, value is dataframe
+            pandas dataframe assigned to keyword argument that produces a dictionary variable.
         """
         for dataset in dataset_dict:
             v = dataset_dict[dataset]
```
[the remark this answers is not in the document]
joemehltretter: @rfratila Updated

rfratila:
@joemehltretter when running `pytest -v` from the tests folder, it can't find the csv files (so they fail). This is something that @nirtiac brought up before but we weren't able to determine the best way to define the path location for the csvs so you can run the tests from any folder. Any ideas?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -4,296 +4,558 @@
...
+                              axis=1).iloc[[2]].values.tolist()[0]
+            xs = torch.tensor(xs, dtype=torch.float)
+            y = self.df[[self.label_column]].iloc[[idx]].values.tolist()[0][0]
+            y = torch.tensor(y, dtype=torch.long)
+            return xs, y
+        else:
+            xs = self.df.iloc[[2]].values.tolist()[0]
+            print(xs)
+            xs = torch.tensor(xs, dtype=torch.float)
+            return xs
```
rfratila: @joemehltretter @nirtiac won't this be a problem in the `forward_pass` function since it's expecting 2 outputs and ignoring the ys?

vulcanai2/datasets/tabulardataset.py
```diff
@@ -4,296 +4,558 @@
...
+        cols: List
+            A list of those columns with threshold percentage null values
         """
-        return list(self.df)
 
-    # TODO: this is really slow make it faster
-    def list_all_numeric_features(self):
+        if threshold >= 1 or threshold <= 0:
+            raise ValueError(
+                "Threshold needs to be a proportion between 0 and 1")
```
rfratila: you should mention that it's exclusive from 0 and 1

vulcanai2/datasets/tabulardataset.py
```diff
@@ -4,296 +4,558 @@
...
-        """
+        strata_field: String
+            name of the examples Field stratified over.
+            Default is 'label' for the conventional label field.
+        random_state: int
+            The random seed used for shuffling.
+
+        Returns
+        -------
+        datasets: Tuple
```
rfratila: you should specify that it's a Tuple of TabularDatasets

joemehltretter:
> 
> 
> @joemehltretter when running `pytest -v` from the tests folder, it can't find the csv files (so they fail). This is something that @nirtiac brought up before but we weren't able to determine the best way to define the path location for the csvs so you can run the tests from any folder. Any ideas?

@rfratila When I run `pytest -v` all tests pass. Which file is it saying it can't locate?

rfratila:
within the `vulcanai2/tests` folder I run `pytest -v` and get `FileNotFoundError: File b'test_data/birthweight_reduced.csv' does not exist`. However, if I run it within `vulcanai2/tests/datasets` it works

rfratila:
Damn this PR is looking awesome well done @nirtiac @joemehltretter ! Might be outside the scope of this but do you think you can patch up the issue with Fashion-MNIST not downloading for the example

joemehltretter:
@rfratila I added functionality that checks where pytest is being ran from; however, I only implemented to be able to successfully run if `pytest -v` is ran in `tests` directory. If it is ran in `vulcanai2` directory is failed. I have another idea that I am going to test to see if it'll work run any Vulcan2 directory.

rfratila:
ok fantastic thanks!

nirtiac:
@joemehltretter love the fix for the path. so simple!

vulcanai2/datasets/tabulardataset.py
```diff
@@ -302,29 +302,21 @@ def create_one_hot_encoding(self, column_name, prefix_sep="@"):
...
     # if a use case presents itself column_name could easily become a list
     def reverse_create_one_hot_encoding(self, column_name=None,
-                                        reverse_all=False, prefix_sep="@"):
+                                        prefix_sep="@"):
         """
         Undo the creation of one-hot encodings, if prefix_sep was used
         to create one-hot encodings and nowhere else. If a column_name is
-        provided, only that column will be reverse-encoded, even if reverse_all
-        is set to true.
+        provided, only that column will be reverse-encoded, otherwise all will
```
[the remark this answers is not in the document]
nirtiac: lol. um. I guess I prefer reverse-encoded because it pairs it more clearly with one_hot_encode.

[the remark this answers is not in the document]
rfratila: ok fair enough. similar to why we don't have thaw instead of unfreeze

vulcanai2/datasets/tabulardataset.py
```diff
@@ -82,8 +83,27 @@ def __init__(self, label_column=None, merge_on_columns=None,
...
 
+        self.seed_value = int(time.time())
+
+        self.random_seeding(self.seed_value)
+
         logger.info(f"You have created a new dataset with {len(self)} rows")
 
+        logger.info(f"The random seed was set to a value of {self.seed_value}")
+
+    def random_seeding(self, seed_value):
```
rfratila: perhaps we should call it `set_global_random_seed()`? or `set_random_seed()`?

rfratila:
> they use numpy https://stackoverflow.com/questions/31057197/should-i-use-random-seed-or-numpy-random-seed-to-control-random-number-gener

perfect no problem then

vulcanai2/tests/datasets/test_utils.py
```diff
@@ -1,26 +1,112 @@
...
+                                            index=[0, 1, 2, 3, 4, 5])
+
+        stitch_dataset_results = stitch_datasets(merge_on_columns=['name', 'dob', 'visit_date'], **my_test_dataset_four)
+
+        # Assert_frame_equal checks order of columns; therefore, sort_index by columns when checking. If dataframes are
+        # same, they should sort the same way.
+        pd.testing.assert_frame_equal(stitch_dataset_results.sort_index(axis=1), df_three_moc_results.sort_index(axis=1),
+                                      check_dtype=False)
+
 if __name__ == '__main__':
```
rfratila: we should remove this since unitest is no longer imported @joemehltretter

nirtiac:
@rfratila I believe that this is finished.

rfratila:
Fantastic @nirtiac @joemehltretter ! I'll merge

PR #102, opened 2018-12-02
TESTS

priyatharsan:
This PR covers all tests that need to be done (including polishing/cleaning existing tests) before the alpha release
**Models**
- [x] models/test_cnn.py
- [x] models/test_dnn.py
- [x] models/layers.py
- [x] models/test_metrics.py
- [x] models/test_utils.py

**Datasets**
- [x] datasets/test_tabulardataset.py
- [x] datasets/test_utils.py

**Viz**
- [ ] plotters/visualizations.py


vulcanai2/tests/models/test_cnn.py
```diff
@@ -81,6 +89,21 @@ def test_forward_class_not_nan(self, cnn_class):
...
+    @pytest.mark.xfail # TODO: Throws RuntimeError: sizes must be non-negative
+    def test_forward_pass_input_net_not_nan(self, 
+                                cnn_class_add_input_network,
+                                cnn_noclass):
+        """Confirm out is non nan."""
+        test_input = torch.ones([1, *cnn_noclass.in_dim])
+        test_dataloader = DataLoader(TensorDataset(test_input, test_input))
+        raw_output = cnn_class_add_input_network.forward_pass(
+            data_loader=test_dataloader,
+            convert_to_class=False)
```
priyatharsan: @rfratila any idea why this is failing? getting `RuntimeError: sizes must be non-negative`

priyatharsan: I doubt this has something to do with casting? not sure

vulcanai2/tests/models/conftest.py
```diff
@@ -0,0 +1,205 @@
...
+            'initializer': None,
+            'bias_init': None,
+            'norm': None,
+            'dropout': 0.5,  # Single value or List
+        },
+        device='cpu'
+    )
+
+@pytest.fixture(scope="module")
+def multi_input_cnn(conv3D_net, multi_input_dnn):
```
rfratila: Can you add another conv net as input?

vulcanai2/tests/models/conftest.py
```diff
@@ -0,0 +1,205 @@
+import pytest
```
rfratila: Can we call this file `test_network_config_fixtures.py` to adhere to the `test_*.py` style of test files?

vulcanai2/tests/models/conftest.py
```diff
@@ -0,0 +1,205 @@
...
+
+@pytest.fixture(scope="module")
+def multi_input_cnn_data(conv3D_net, multi_input_dnn_data):
+    return MultiDataset([
+        (TensorDataset(torch.ones([10, *conv3D_net.in_dim])), True, False),
+        multi_input_dnn_data
+    ])
+
+@pytest.fixture(scope="module")
+def cnn_noclass():
```
rfratila: I don't know if we need a conv2D_net AND a cnn_noclass. perhaps we can just pick one? lowers the number of fixtures to create?

vulcanai2/tests/models/test_cnn.py
```diff
@@ -1,66 +1,44 @@
...
-                    {
-                        "in_channels": 16,
-                        "out_channels": 1,
-                        "kernel_size": (5, 5),
-                        "stride": 2,
-                        "padding": 2
-                    }]
-            }
-        )
+    def test_forward_multi_input_ConvNet(self, multi_input_cnn,
```
rfratila: you don't need to include ConvNet in the title since it already exists within the class `TestConvNet`

vulcanai2/tests/models/conftest.py
```diff
@@ -0,0 +1,205 @@
...
+        },
+        num_classes=3
+    )
+
+@pytest.fixture(scope="module")
+def cnn_class_add_input_network(cnn_noclass, cnn_class):
+    """Create ConvNet with input_network added via
+    add_input_network and has a prediction layer."""
+    net = cnn_class
+    net.add_input_network(cnn_noclass)
```
rfratila: I think it adds the cnn_noclass to cnn_class which is why you might be getting the error when testing the forward on this function

vulcanai2/tests/models/test_cnn.py
rfratila: see my comment about the fixture about this error

vulcanai2/tests/models/conftest.py
priyatharsan: I think to have a shared fixtures, it must be named as `conftest.py` : https://docs.pytest.org/en/latest/writing_plugins.html#conftest-py-plugins

vulcanai2/tests/models/conftest.py
priyatharsan: 👍

vulcanai2/tests/models/conftest.py
rfratila: ah ok no worries then. thanks for the link!

vulcanai2/tests/models/conftest.py
priyatharsan: hmm I'm not quite sure if I understood `add_input_network`, lets say if a main network does not have any `input_networks` so how do I add a new input_network? I pushed a redefined example in the next commit. Let me know what am I missing.

rfratila: AHH I know the problem! So cnn_class has a dense output which you are trying to feed into a conv net. It's an edge case that I let fail since I think it doesn't make much sense for that sort of scenario (e.g. fully connected -> cnn). If you disagree, then it would not be a difficult fix; you would just need to pass a cast_shape of the receiving network to cast the dense input to that. @priyatharsan @joemehltretter @nirtiac I'd like to get your input on this.

priyatharsan: I think its good to have that ability to convert FC to CNN, as we already have the ability to convert [FC + some other conv input] to a CNN.

priyatharsan: However, I'm confused in this case `cnn_class.add_input_network(cnn_noclass)`, I'm adding `cnn_noclass` (which has a conv output) as an input_network to the `cnn_class` right?

rfratila: @priyatharsan the issue is with information loss. dense + conv 1D inputted to conv 2D doesn't make much sense since you're not getting any additional benefit of padding a bunch of zeros in the extra dimension. Really the input network should have a size of the max(list_input_networks_output_size) which is what is happening now. Similarly, FC -> Conv doesn't really add much benefit since there's no spatial information in FC. Does that make sense? Regarding debugging, I would see what the cnn_noclass output shape is

vulcanai2/models/basenetwork.py
```diff
@@ -160,6 +160,8 @@ def add_input_network(self, in_network):
         None
 
         """
+        # TODO: I think we need to re-create/re-initalize the basenetwork when calling this function since 
+        # self.network has no idea how the new input shape will look like after adding a new input_network?
```
rfratila: That's not true. That's why we run `self.in_dim = self._get_in_dim()`

vulcanai2/tests/models/test_cnn.py
```diff
@@ -1,114 +1,140 @@
...
+                             multi_input_cnn_data):
+        """Test for fit function"""
+        test_train = Subset(multi_input_cnn_data, 
+                            range(len(multi_input_cnn_data)//2))
+        test_val = Subset(multi_input_cnn_data, 
+                          range(len(multi_input_cnn_data)//2, 
+                                    len(multi_input_cnn_data)))
+        test_train_loader = DataLoader(test_train, batch_size=2)
+        test_val_loader = DataLoader(test_val, batch_size=2)
+        multi_input_cnn.fit(test_train_loader, test_val_loader, 1)
```
rfratila: @priyatharsan would you be able to assert that the weights (I.e. parameters) have changed? you can probably compare parts of the `state_dict`

priyatharsan: sure I will take a look into it tonight.

vulcanai2/tests/models/test_cnn.py
```diff
@@ -1,10 +1,13 @@
 """Test all ConvNet capabilities."""
 import numpy as np
 import pytest
+import copy
 
 import torch
 import torch.nn as nn
 from torch.utils.data import DataLoader, Subset, TensorDataset
+from torch.autograd.gradcheck import gradcheck
```
rfratila: You don't seem to be using `gradcheck`?

vulcanai2/tests/models/test_cnn.py
```diff
@@ -136,5 +139,19 @@ def test_fit_multi_input(self, multi_input_cnn,
                           range(len(multi_input_cnn_data)//2, 
                                     len(multi_input_cnn_data)))
         test_train_loader = DataLoader(test_train, batch_size=2)
-        test_val_loader = DataLoader(test_val, batch_size=2)
-        multi_input_cnn.fit(test_train_loader, test_val_loader, 1)
\ No newline at end of file
+        test_val_loader = DataLoader(test_val, batch_size=2)     
+        
+        init_weights = copy.deepcopy(multi_input_cnn.network[0]._kernel.weight.data)
```
rfratila: you don't need to use `.data` as it's deprecated; `.detach()` should be enough or just extracting `.weight`

vulcanai2/tests/models/test_cnn.py
```diff
@@ -136,5 +139,19 @@ def test_fit_multi_input(self, multi_input_cnn,
...
+        multi_input_cnn_no_fit = copy.deepcopy(multi_input_cnn)
+        try:
+            multi_input_cnn.fit(test_train_loader, test_val_loader, 2)
+        except RuntimeError:
+            print("The network multi_input_cnn failed to train.")
+        finally:
+            trained_weights = multi_input_cnn.network[0]._kernel.weight.data.cpu()
+            
+            # Sanity check if the network parameters are training
+            assert (torch.equal(init_weights.cpu(), trained_weights.cpu()) is False)
```
rfratila: I don't think you need to specify cpu on this

vulcanai2/tests/models/test_cnn.py
```diff
@@ -136,5 +139,19 @@ def test_fit_multi_input(self, multi_input_cnn,
...
-        multi_input_cnn.fit(test_train_loader, test_val_loader, 1)
\ No newline at end of file
+        test_val_loader = DataLoader(test_val, batch_size=2)     
+        
+        init_weights = copy.deepcopy(multi_input_cnn.network[0]._kernel.weight.data)
+        multi_input_cnn_no_fit = copy.deepcopy(multi_input_cnn)
+        try:
+            multi_input_cnn.fit(test_train_loader, test_val_loader, 2)
+        except RuntimeError:
+            print("The network multi_input_cnn failed to train.")
```
rfratila: please use the logger instead of printing

vulcanai2/tests/models/test_cnn.py
priyatharsan: oops yeah I will removed. Grad check isn't compatible with our data types.

vulcanai2/tests/models/test_cnn.py
priyatharsan: It throws error if I don't, since the model is in cuda in `test_forward_multi_input`

vulcanai2/tests/models/test_cnn.py
rfratila: why not? seems like something that should be used

vulcanai2/tests/models/test_cnn.py
rfratila: have you ran all the tests on just CPU to confirm there's no issues?

vulcanai2/tests/models/test_cnn.py
priyatharsan: cause our data is formatted as a list. Take a look at here: https://github.com/pytorch/pytorch/blob/331c4b5b4d600049fbd5a6503c9cf5f1009214c5/torch/autograd/gradcheck.py#L205

priyatharsan: I tried hacking it, but took a while and couldn't get through it sadly

vulcanai2/tests/models/test_cnn.py
priyatharsan: 👍

vulcanai2/tests/models/test_cnn.py
rfratila: looking at that code it shouldn't be an issue since it will just return the list as a tuple

vulcanai2/tests/models/test_cnn.py
priyatharsan: yeah but by default it has to be fed as a tuple. Take a look at here as well: https://discuss.pytorch.org/t/interpreting-gradcheck-errors/16239/11

vulcanai2/tests/models/test_cnn.py
priyatharsan: yeah but by default it has to be fed as a tuple. Take a look at here as well: https://discuss.pytorch.org/t/interpreting-gradcheck-errors/16239/11

>

vulcanai2/tests/models/test_cnn.py
priyatharsan: are you getting any errors? Not sure how do I make pytorch to ignore my gpu rather than uninstalling CUDA from my machine. Manually switch of device tests are already carried out in `test_device.py` I don't get any errors there.

vulcanai2/tests/models/test_cnn.py
priyatharsan: Here: https://github.com/pytorch/pytorch/blob/331c4b5b4d600049fbd5a6503c9cf5f1009214c5/torch/autograd/gradcheck.py#L152-L158 , if passed as list will get converted into tuple. And if I pass as tuple(list(data)) then it throws error for requires_grad=True (https://github.com/pytorch/pytorch/blob/331c4b5b4d600049fbd5a6503c9cf5f1009214c5/torch/autograd/gradcheck.py#L223)

vulcanai2/tests/models/test_cnn.py
rfratila: tests pass for just CPU can confirm

vulcanai2/tests/models/test_cnn.py
rfratila: weird. perhaps the state_dict is getting buried unnecessarily?

priyatharsan: no it didn't reach the point where forward `(model(date)` or `model.forward(date))` is happening, it just failing around the typecheck within the gradcheck

rfratila: it should be working. I'd like you to look a little more into it if there's any major issues since this should work no problem

priyatharsan: I think you didn't get what my point is, so it checks firstly if the `inputs` is a list or tuple. Either case it converts to `tuple` and then for each item in the` tuple` it checks for `requires_grad=True` our case will definitely fail, especially if the network is a multi-input_network with one of the `input_network` itself is a multi-input (where our data-set will be a `list` of `list`s). So list doesn't have requires_grad.

vulcanai2/models/basenetwork.py
```diff
@@ -141,7 +140,11 @@ def __init__(self, name, config, in_dim=None, save_path=None,
         self.out_dim = self._get_out_dim()
 
         self.device = device
+        self._optim_spec = optim_spec
         self._criter_spec = criter_spec
+        self.optim = None
+        self.criterion = None
+        self._net_spec_initialzed = False
```
rfratila: @priyatharsan why are you creating this class attribute?

vulcanai2/models/basenetwork.py
```diff
@@ -489,6 +492,7 @@ def _init_criterion(criterion_spec):
         return criterion_spec
 
     def _init_trainer(self):
+        self._net_spec_initialzed = True
```
rfratila: If you want to determine if the optimizer was build you can simply do `if self.optim` and as long as it's None it's there. We don't need a class attribute just for this

vulcanai2/tests/models/test_cnn.py
rfratila: you just have to write `logger.error("string")`. see the other files for more examples. I'd like this to be done before merging

vulcanai2/tests/models/test_cnn.py
```diff
@@ -131,27 +144,64 @@ def test_unfreeze_noclass(self, conv3D_net):
...
\ No newline at end of file
+                                is False)
+    
+    def test_params_multi_input(self, multi_input_cnn,
+                                multi_input_train_loader,
+                                multi_input_test_loader):
+        """Test for change in network params/specifications"""
+        
+        test_net1 = copy.deepcopy(multi_input_cnn)
+        test_net2 = pickle.loads(pickle.dumps(multi_input_cnn))
```
rfratila: I'm not sure why you are doing this. Would you be able to clarify?

vulcanai2/tests/models/test_cnn.py
```diff
@@ -204,4 +204,14 @@ def test_params_multi_input(self, multi_input_cnn,
...
                                         test_net1.optim.param_groups[0]['params']):
                 assert param is opt_param
-        
\ No newline at end of file
+        
+        # Check the params after saving loaading
+        test_net2.save_model()
+        save_path = test_net2.save_path
+        loaded_test_net2 = BaseNetwork.load_model(load_path=save_path)
+        # Warning: Repeated run of this test will resulting dumping
```
rfratila: No you need to delete the saved model instantly. Nothing should persist after running the test. See the tabular dataset csv saving test for an example of saving, loading and deleting.

vulcanai2/tests/models/test_cnn.py
```diff
@@ -204,4 +204,14 @@ def test_params_multi_input(self, multi_input_cnn,
...
+        # the saved model into the storage
+        load_params = [torch.allclose(param1, param2)
```
rfratila: I'm not exactly sure what `torch.allclose` does but you should also check that the parameters are equivalent (unless that's what torch.allclose does but then your previous test of fitting 2 separate models and checking allclose with the original model doesn't make too much sense since it's not guaranteed that the weights will be the same after 2 epochs of training)

.gitignore
```diff
@@ -6,6 +6,7 @@
 data
 runs
 examples/saved_models
+vulcanai2/tests/models/saved_models/
```
rfratila: since you will delete the file after saving, you won't need this anymore

vulcanai2/tests/models/test_cnn.py
priyatharsan: just wanted to make sure if copying/duplicating the models via `deepcopy` or `pickle` does copy all the parameters? If you ask about a use-case for copying: I thought about what if we have a certain network module within a multi-input network, make a copy of it and insert it somewhere else within the same multi-input network using `add_input_network`. Not sure how feasible that is.

vulcanai2/tests/models/test_cnn.py
priyatharsan: `torch.allclose` does check the parameters element-wise: https://pytorch.org/docs/stable/torch.html#torch.allclose

>checking allclose with the original model doesn't make too much sense since it's not guaranteed that the weights will be the same after 2 epochs of training
Right now the parameters are not the same for when training 2 epochs and the test is passing. Not sure if I quite understood what you mean it does not makes sense

.gitignore
priyatharsan: 👍

vulcanai2/tests/models/test_dnn.py
```diff
@@ -97,3 +136,79 @@ def test_add_input_network(self, dnn_noclass, dnn_class):
...
+
+    # TODO: Failing fit! Fix the test
+    def test_fit_multi_input(self, multi_input_dnn,
+                             multi_input_train_loader,
+                             multi_input_test_loader):
+        """Test for fit function"""        
+        init_weights = copy.deepcopy(multi_input_dnn.network[0]._kernel.weight.detach())
+        multi_input_dnn_no_fit = copy.deepcopy(multi_input_dnn)
+        parameters1 = multi_input_dnn_no_fit.parameters()
+        try:
```
priyatharsan: Its failing the test here. probably it has something to do with how the dataloader is created. Any idea?

vulcanai2/tests/models/test_dnn.py
```diff
@@ -37,13 +51,38 @@ def test_init(self, dnn_noclass):
...
+    # TODO: Failing! Fix the test
+    def test_forward_multi_input_dnn_add_input_network(self, 
+                                multi_input_dnn_add_input_network):
+        """Test Forward of Multi Input DenseNet where input_networks
+        added via add_input_network"""
+        out = multi_input_dnn_add_input_network([
+                            torch.ones([10, 1, 28]),
+                            torch.ones([10, 1, 28, 28]),
+                            torch.ones([10, 200])
+                        ])
```
priyatharsan: @rfratila and this one fails. I think we need to replace the first layer of the basenetworks's`.network` with the updated `self._get_in_dim()` after adding the `input_network` via `add_input_network`

priyatharsan: I believe this would solve the issue we had with the cnn as well earlier

vulcanai2/models/dnn.py
```diff
@@ -157,8 +157,10 @@ def _create_network(self, **kwargs):
         self.network = nn.Sequential(dense_layers)
 
         if self._num_classes:
-            self.network.add_module(
-                'flatten', FlattenUnit())
+            if self._get_out_dim is not None and \
```
rfratila: Why add this condition?

vulcanai2/tests/models/test_cnn.py
rfratila: aah I see. thanks for the clarification!

vulcanai2/tests/models/test_dnn.py
rfratila: For now @priyatharsan turn `add_input_network` into a class private `_add_input_network` function to avoid adding a network after it's been created until we come up with a better solution. For now it's simpler if an input_network is specified at construction and not afterworlds. We won't run into these issues

vulcanai2/tests/models/test_dnn.py
rfratila: You'll have to be more specific about the error your getting please

vulcanai2/tests/models/test_dnn.py
```diff
@@ -86,3 +136,79 @@ def test_add_input_network(self, dnn_noclass, dnn_class):
...
+            multi_input_dnn.fit(multi_input_train_loader, 
+                                multi_input_test_loader, 2)
+        except RuntimeError:
+            logger.error("The network multi_input_dnn failed to train.")
+        finally:
+            parameters2 = multi_input_dnn.parameters()
+            trained_weights = multi_input_dnn.network[0]._kernel.weight.detach().cpu()
+            
+            # Sanity check if the network parameters are training
+            assert (torch.equal(init_weights.cpu(), trained_weights.cpu()) is False)
```
rfratila: do you need to call `.cpu()` twice?

vulcanai2/tests/models/test_cnn.py
rfratila: sure. you can do that as long as the dimensions of the inputs and outputs agree. it just becomes another independent module if you `deepcopy` it.

vulcanai2/tests/models/test_dnn.py
priyatharsan: that's a mistake. I will remove the repeated one.

vulcanai2/tests/models/test_cnn.py
rfratila: any updates?

vulcanai2/tests/models/test_dnn.py
rfratila: do not replace any layer.

vulcanai2/tests/models/test_cnn.py
priyatharsan: no, unfortunately I'm still stuck there.

rfratila: it's a list with embedded lists so loop through and recursively go inside embedded lists and apply this check then if it's not a list.

vulcanai2/tests/models/test_cnn.py
priyatharsan: I can remove `deepcopy` testing in that case. Just wanted to prove the users that it is possible to use both copying techniques.

vulcanai2/models/dnn.py
priyatharsan: it looks like `flatten` layer gets added for all use-cases. We wouldn't need flatten layer when all our `input_networks` are `DenseNet`s

vulcanai2/tests/models/test_cnn.py
rfratila: might take up a lot of memory to copy several times

vulcanai2/tests/models/test_cnn.py
priyatharsan: This error seems to be common out there, but couldn't able to find a way to get through it. Even the post ⬆️ I shared is related to that and `.double()` supposed to fix it, but it didn't in this case.

vulcanai2/models/dnn.py
rfratila: might as well check if the input network `isinstance` of a `DenseNet` no? it's what you're indirectly checking

vulcanai2/tests/models/test_cnn.py
rfratila: Im not sure if `conv1D_net.double()` is enough to cast the entire internal network to double as well. perhaps that can be the issue? if this is causing too many issues and it's a number casting issue then perhaps we should remove this check. but try to first thing I suggested first

vulcanai2/models/dnn.py
priyatharsan: hmm I could add a check for if all the input_networks are `DenseNet`s but why not just check the shape of the input_tensor after merging the networks? I feel thats much simpler, no?

vulcanai2/tests/models/test_cnn.py
priyatharsan: >but try to first thing I suggested first
 if you mean
>so loop through and recursively go inside embedded lists and apply this check then if it's not a list.
that's what this is: 
>Here is the snippet I tried with:
`gradcheck(conv1D_net.double(), (torch.ones([10, 1, 28], requires_grad=True).double(),))`

isn't it?

rfratila: I mean go through each layer in the network and cast it to double precision instead of the BaseNetwork object.

vulcanai2/models/dnn.py
rfratila: 1) you can't check if a function is not None i.e. `if self._get_out_dim is not None`, I think you mean to call the actual function instead of the reference to it. I'm not sure which is slower, to have a redundant flat call or check a condition. would you be able to speed test it? I think that just having the redundant flatten layer is just as fast

vulcanai2/tests/models/test_cnn.py
priyatharsan: removed it now

vulcanai2/models/dnn.py
priyatharsan: >I think that just having the redundant flatten layer is just as fast

Probably, but this adds a flatten layer unnecessarily to the graph which might lead to confusions, don't you think so?

priyatharsan: I will take a look into speed testing

vulcanai2/tests/models/test_cnn.py
priyatharsan: @rfratila I think I found the culprit, its failing [here](https://github.com/Aifred-Health/Vulcan2/blob/test_branch/vulcanai2/tests/models/test_cnn.py#L108), could this be anything to do with the order of `activation` and `pool` ?

vulcanai2/models/dnn.py
priyatharsan: not much of a difference in speed. The shape checking was approx faster by .05 seconds in tests, I switched to instance checking of `input_networks` as you suggested 👍

vulcanai2/models/dnn.py
```diff
@@ -157,8 +157,8 @@ def _create_network(self, **kwargs):
         self.network = nn.Sequential(dense_layers)
 
         if self._num_classes:
-            if self._get_out_dim is not None and \
-                len(self._get_out_dim()) >= 2:
+            if self.input_networks is not None and\
```
rfratila: Actually I don't think it is needed to add flatten since it comes after a dense network always. the input networks flow through the current network first and thus won't need to be flattened

vulcanai2/tests/models/conftest.py
```diff
@@ -223,13 +223,13 @@ def multi_input_cnn_add_input_network(conv1D_net, conv2D_net,
         },
         device='cpu'
     )
-    net.add_input_network(conv3D_net)
```
rfratila: you shouldn't add an input after initiating the network hence why it is a private function

priyatharsan: makes sense! yeah

vulcanai2/tests/models/test_cnn.py
rfratila: I don't think the order is the problem since that's quite standard. Don't you need to run inp through the sublayer before doing a gradcheck? also you should use a different var name then net while you go through the sublayers of a network.

priyatharsan: > Don't you need to run inp through the sublayer

ofcourse I do run the `inp` to the layer, please check: https://github.com/Aifred-Health/Vulcan2/blob/test_branch/vulcanai2/tests/models/test_cnn.py#L99

rfratila: yes you run it through but don't you need to do it before you do a gradcheck with that sublayer?

priyatharsan: no gradcheck already does a forward inside: https://github.com/pytorch/pytorch/blob/331c4b5b4d600049fbd5a6503c9cf5f1009214c5/torch/autograd/gradcheck.py#L228

priyatharsan: which makes me realize now:
```
assert gradcheck(net._kernel.double(), (inp.double(),))
inp = net._kernel.double()(inp.double())
```
doesn't makes sense as well I think.

rfratila: try doing one layer at a time instead of the sublayers

priyatharsan: I tried the before, it was breaking that's why I went sublayer level. I'd appreciate if I can get some help. I've been going back and forth on this.

rfratila: I don't think `gradcheck` is the best thing to use here: https://discuss.pytorch.org/t/check-gradient-flow-in-network/15063

vulcanai2/tests/models/test_dnn.py
priyatharsan: resolved it in the latest commit

vulcanai2/tests/models/test_cnn.py
priyatharsan: no wonder I was breaking my head on that lol

priyatharsan:
I think this PR is done, unless you have anything else in mind @rfratila

rfratila:
You don't test the utils @priyatharsan yet you checked it off. Don't worry about the selu specific functions since those are already tested

priyatharsan:
oh I thought it was covered in https://github.com/Aifred-Health/Vulcan2/pull/101. Okay I will work on that.

rfratila:
Also @priyatharsan you didn't add layer specific tests yet. Please make sure that creating those layers with a dict-as-keywords methodology works and returns errors if the wrong key word was passed

rfratila:
> oh I thought it was covered in #101. Okay I will work on that.

@priyatharsan that was for dataset utils not the model utils

vulcanai2/models/layers.py
```diff
@@ -43,7 +43,6 @@ def __init__(self, initializer=None, bias_init=None,
         self.out_bound_layers = []
 
         self._kernel = None
-        self.norm = norm
```
rfratila: why are you removing this?

vulcanai2/tests/models/test_layers.py
```diff
@@ -1,42 +1,92 @@
...
 
 class TestBaseUnit:
     @pytest.fixture
     def baseunit(self): 
         return BaseUnit()
 
+    def test_init(self, baseunit):
+        """Initialization Test of a BaseUnit object"""
+        assert isinstance(baseunit, nn.Sequential)
+        assert hasattr(baseunit, 'initializer')
```
rfratila: it's called `weight_init` now

rfratila: could you add a quick `TODO` so we don't forget? thanks!

vulcanai2/tests/models/test_layers.py
```diff
@@ -48,6 +48,27 @@ def test_init(self, denseunit):
...
+                    norm='batch',
+                    activation=nn.ReLU(),
+                    dropout=0.5
+                )
+    
+    def test_create_denseunit(self, test_denseunit_parameters):
+        """Check if passing wrong parameters raises TypeError"""
+        with pytest.raises(TypeError) as e:
+            DenseUnit(**test_denseunit_parameters)
+            assert e.message == "__init__() got an unexpected \
```
rfratila: this assertion should be outside of the `with` statement

vulcanai2/tests/models/test_utils.py
```diff
@@ -0,0 +1,68 @@
...
+    decimals"""
+    test_list = [0.83754245, 0.13249807]
+    out = round_list(test_list, decimals=3)
+    assert len(str(out[0]).split(".")[1]) == 3
+
+def test_get_one_hot():
+    """Test for get_one_hot"""
+    test_inp = np.array([0,1,2,3,4,3,2,1,0])
+    n_classes = 5
+    assert get_one_hot(test_inp).shape == (len(test_inp),
```
rfratila: would you be able to actually handcode the one-hot matrix and see if it returns the same thing? might be the simplest and most robust test for this.

vulcanai2/tests/models/test_utils.py
```diff
@@ -0,0 +1,68 @@
...
+                                           n_classes)
+
+def test_pad():
```
rfratila: perhaps you should also try with at least one other 2d or 3d sized tensor?

vulcanai2/tests/models/test_utils.py
```diff
@@ -0,0 +1,68 @@
...
+    test_nchannels = 2
+    test_nfeatures = 50
+    padded_shape = 250
+    test_inp = torch.randn([test_nchannels, test_nfeatures])
+    padded_tensor = pad(tensor=test_inp, padded_shape=[padded_shape])
+    assert padded_tensor.shape == (test_nchannels, padded_shape)
+    assert padded_tensor.nonzero().size(0) == test_nchannels*\
+                                              test_nfeatures
+
+def test_set_tensor_device():
```
rfratila: you could probably add another pytest skip here since it deals with cuda device changes

vulcanai2/tests/models/test_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        
+        # List of Tensors
+        test_tensorlist_cpu = [torch.randn([1, 2]).cpu(),
+                           torch.randn([2, 3]).cpu(),
+                           torch.randn([3, 4]).cpu()]
+        test_tensorlist_cuda = set_tensor_device(test_tensorlist_cpu, device='cuda:0')
+        for t in test_tensorlist_cuda:
+            assert str(t.device) == 'cuda:0'
+
+def test_master_device_setter(multi_input_cnn):
```
rfratila: same comment about skipping with pytest since cuda is involved

vulcanai2/tests/models/test_layers.py
```diff
@@ -70,6 +91,27 @@ def convunit(self):
...
+                    norm='batch',
+                    activation=nn.ReLU(),
+                    dropout=0.5
+                )
+    
+    def test_create_convunit(self, test_convunit_parameters):
+        """Check if passing wrong parameters raises TypeError"""
+        with pytest.raises(TypeError) as e:
+            DenseUnit(**test_convunit_parameters)
+            assert e.message == "__init__() got an unexpected \
```
rfratila: same comment about moving the assertion outside the with statement

vulcanai2/models/layers.py
rfratila: ah ok. good catch!

vulcanai2/tests/models/test_layers.py
priyatharsan: 👍

vulcanai2/tests/models/test_layers.py
priyatharsan: 👍

vulcanai2/tests/models/test_utils.py
priyatharsan: 👍

vulcanai2/tests/models/test_utils.py
priyatharsan: 👍

vulcanai2/tests/models/test_utils.py
priyatharsan: done

vulcanai2/tests/models/test_layers.py
```diff
@@ -78,8 +78,8 @@ def test_create_denseunit(self, test_denseunit_parameters):
         """Check if passing wrong parameters raises TypeError."""
         with pytest.raises(TypeError) as e:
             DenseUnit(**test_denseunit_parameters)
-            assert e.message == "__init__() got an unexpected \
-                                keyword argument 'in_channels'"
+        assert e.message == "__init__() got an unexpected \
```
rfratila: I get the following error ```assert e.message == "__init__() got an unexpected \
                                keyword argument 'in_channels'"
E       AttributeError: 'ExceptionInfo' object has no attribute 'message'```.

priyatharsan: Hmm I just tested now, I'm getting the same. Not sure why the error message changed, but this could be because from the merge commits I guess

priyatharsan: >~but this could be because from the merge commits I guess~

Its actually cuz I moved the assert out of `with`

priyatharsan:
Thanks for the changes @rfratila , also @joemehltretter would you like to continue with visualization tests in this PR?

rfratila:
Actually @joemehltretter you should add the visualization tests to the remote_plotting PR since this one is getting pretty big. @nirtiac please create a new PR for all the metrics tests. if either of you @joemehltretter or @nirtiac had some free time to review this PR please let me know otherwise it's time to merge it.

joemehltretter:
@rfratila So put the tests in the remote_server_plotting or create a new one?

rfratila:
@joemehltretter you can put it in the `remote_server_plotting` PR

vulcanai2/tests/models/test_cnn.py
```diff
@@ -1,114 +1,204 @@
...
+        assert isinstance(multi_input_cnn.input_networks, nn.ModuleDict)
+        assert len(list(multi_input_cnn.input_networks)) == 3
+        assert all(multi_input_cnn._get_max_incoming_spatial_dims() == (8, 8, 8))
+        assert multi_input_cnn._merge_input_network_outputs([
+                torch.ones([10, 1, 28, 28]),
+                torch.ones([10, 1, 28, 28, 28]),
+                torch.ones(10, *multi_input_cnn.
+                           input_networks['multi_input_dnn'].out_dim)
+            ]).shape == (10, 3, 8, 8, 8)
+        test_net = pickle.loads(pickle.dumps(multi_input_cnn))
```
nirtiac: ?? Can someone explain why this is here?

vulcanai2/tests/models/test_cnn.py
```diff
@@ -1,114 +1,204 @@
...
+    def test_forward(self, conv1D_net):
+        """Test Forward of ConvNet."""
+        out = conv1D_net(torch.ones([10, *conv1D_net.in_dim]))
+        assert out.shape == (10, 64, 1)
+
+    def test_forward_multi_input(self, multi_input_cnn, conv1D_net):
+        """Test Forward of Multi Input ConvNet."""
+        master_device_setter(multi_input_cnn, 'cpu')
+        input_tensor = [
+            torch.ones([10, 1, 28, 28]),
```
nirtiac: I'm not completely confident we should be using torch.ones in all of our tests? An alternate might be         np.random.randint(0, 9, size=num_items)

vulcanai2/tests/models/test_cnn.py
```diff
@@ -1,114 +1,204 @@
...
+
+        # Check the parameters change after copy and fit
+        test_net.fit(
+            multi_input_cnn_train_loader,
+            multi_input_cnn_test_loader,
+            2)
+        close_params = [not torch.allclose(param1, param2)
+                        for param1, param2 in zip(multi_input_cnn.parameters(),
+                                                  test_net.parameters())]
+        assert all(close_params)
```
nirtiac: nice, I like this

vulcanai2/tests/models/test_cnn.py
rfratila: the emphasis was on using more than just ones. sure we can use torch.randn but we just need to make sure about making deterministic test cases

priyatharsan: not sure I got your point... can you point me out what you mean by deterministic test cases please?

vulcanai2/tests/models/test_cnn.py
nirtiac: why this instead of deepcopy? I'm pretty sure it would be faster.

vulcanai2/tests/models/test_cnn.py
nirtiac: I think Robert was just pointing out that we need to use "random" numbers, or something better than just ones, but they need to return the same result every time. To do that we need to set a random seed, so there's two things that need to be changed.

vulcanai2/tests/models/test_dnn.py
```diff
@@ -81,8 +129,70 @@ def test_unfreeze_noclass(self, dnn_noclass):
...
+                multi_input_dnn_test_loader,
+                2)
+        except RuntimeError:
+            logger.error("The network multi_input_dnn_class failed to train.")
+        finally:
+            parameters2 = multi_input_dnn_class.parameters()
+            trained_weights = multi_input_dnn_class.network[0]._kernel.weight.detach()
+            
+            # Sanity check if the network parameters are training
+            assert not (torch.equal(init_weights.cpu(), trained_weights.cpu()))
```
nirtiac: does this work with float values?

vulcanai2/models/basenetwork.py
```diff
@@ -141,9 +140,12 @@ def __init__(self, name, config, in_dim=None, save_path=None,
         self.out_dim = self._get_out_dim()
 
         self.device = device
+        self._optim_spec = optim_spec
         self._criter_spec = criter_spec
+        self.optim = None
```
nirtiac: This was much debated before, you're just adding it back in so that all class parameters are included in init? nothing else changes/needs to change?

vulcanai2/tests/models/test_cnn.py
priyatharsan: good to know. didn't know deepcopy is faster. I just tested it. Thanks.. I changed it to deepcopy now

vulcanai2/tests/models/test_cnn.py
priyatharsan: @rfratila @nirtiac would any of you be interested in trying it out with the deterministic approach? I'm not clear with the idea since I didnt know about it.

vulcanai2/tests/models/test_cnn.py
```diff
@@ -1,114 +1,204 @@
...
-    def test_add_input_network(self, cnn_noclass, cnn_class):
-        """Test add input Network functionality."""
-        cnn_class.add_input_network(cnn_noclass)
-        assert cnn_class.input_networks[cnn_noclass.name] is cnn_noclass
-        assert cnn_class.in_dim == cnn_noclass.out_dim
+    def test_fit_multi_input(self, multi_input_cnn,
+                             multi_input_cnn_train_loader,
+                             multi_input_cnn_test_loader):
+        """Test for fit function."""
+        init_weights = pickle.loads(pickle.dumps(
```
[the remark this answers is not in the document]
priyatharsan: why do you want it to comment? its copying the weights to compare with if they are training after fit.

vulcanai2/tests/models/test_device.py
```diff
@@ -19,182 +18,55 @@
...
-        assert multi_net.device == torch.device(type='cuda', index=0)
-        assert multi_net.input_networks['conv3D_net']\
+        assert hasattr(multi_input_cnn, 'device')
+        master_device_setter(multi_input_cnn, 'cuda:0')
+        assert multi_input_cnn.device == torch.device(type='cuda', index=0)
+        assert multi_input_cnn.input_networks['conv3D_net']\
             .device == torch.device(type='cuda', index=0)
-        assert multi_net.input_networks['dense_net']\
+        assert multi_input_cnn.input_networks['multi_input_dnn']\
             .device == torch.device(type='cuda', index=0)
```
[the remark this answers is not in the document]
priyatharsan: if the running machine has more than one gpu (in our case, the ml server), it takes the first gpu to test.

vulcanai2/tests/models/test_dnn.py
```diff
@@ -1,40 +1,88 @@
...
-from vulcanai2.models.dnn import DenseNet
-from torch.utils.data import TensorDataset, DataLoader
+import torch.nn as nn
+from torch.utils.data import DataLoader, Subset, TensorDataset
+
+from vulcanai2.models import BaseNetwork
+from vulcanai2.models.dnn import DenseNet, DenseNetConfig
+from vulcanai2.models.utils import master_device_setter
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
priyatharsan: yep we do need it: https://github.com/Aifred-Health/Vulcan2/blob/test_branch/vulcanai2/tests/models/test_dnn.py#L146

vulcanai2/tests/models/test_dnn.py
priyatharsan: yeah the weights are float values.

vulcanai2/tests/models/test_cnn.py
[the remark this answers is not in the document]
nirtiac: just cause it wasn't, like easily readable, but whatever.

vulcanai2/tests/models/test_cnn.py
rfratila: @priyatharsan she described specifying a random seed and then using randint to randomly create the class targets

vulcanai2/tests/models/test_cnn.py
```diff
@@ -1,114 +1,204 @@
...
+    def test_freeze_noclass(self, conv3D_net):
         """Test intermediate network freezing."""
-        cnn_noclass.freeze(apply_inputs=False)
-        for params in cnn_noclass.network.parameters():
+        conv3D_net.freeze(apply_inputs=False)
+        for params in conv3D_net.network.parameters():
             assert params.requires_grad is False
 
-    def test_unfreeze_noclass(self, cnn_noclass):
+    def test_unfreeze_noclass(self, conv3D_net):
```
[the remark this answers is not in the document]
rfratila: they're really similar but it's just to see that appending the class layer works with freezing the network. A bit redundant? yeah but I think it should stay. can never be too careful :)

nirtiac:
Updated to use random integers from 0-9 instead of just ones. Set a random seed, which isn't necessary for now but will be should we want to check outputs in the future. Am a bit concerned about the case where someone loads Vulcan, then also runs tests, changing the global random seed values. This case is admittedly rare.

vulcanai2/tests/models/test_dnn.py
nirtiac: @priyatharsan @rfratila  Yeah I know they're float values. My point is that generally equality functions don't work with float values; you have to check that they're approximately equal or equal up to a certain number of decimal points. Looking at the implementation of the function https://github.com/torch/torch7/pull/470 I'm not sure that it actually fulfills what we need... it may just be checking the size, and thus this condition will always be True.

rfratila: @nirtiac yes this is a good point. @joemehltretter uses an approximate comparison with bounds for testing the selu weight inits that we can use here. Could you adapt some of that code and apply it here?

nirtiac: Or use np.testing.assert_almost_equal(res, target_res)

rfratila: yeah I think that's what he uses

PR #106, opened 2018-12-23
Metric tests

nirtiac:
I believe the extra commits should disappear after the merge of tests. 

Three things for discussion
1. I still get warnings, because the model fails at predicting for my fake data. 

2. I have code that is duplicated in many but not all methods, do you think I should make this code into some sort of utility function? seemed almost as much trouble to call that function, really but might be nicer for changes in the future? 

3. I did some manual verification of the easier metrics (aka did the calc again), but it'd be nice if we could run all our metrics on some sort of standard dataset and compare to others' metric results. Does anyone know of such a dataset?


nirtiac:
@rfratila I could just put these commits into the test branch...

rfratila:
@nirtiac in response to your questions.

1. I think if we trained it for even 1 epoch you would get a more diverse set of predictions instead of always the same class OR we can just create random prediction vectors to compare to the targets and use that instead?
2. How much additional work would it be to create these utility functions? it would be ideal but I don't want to impose...
3. im not familiar with such a dataset. perhaps @joemehltretter or @priyatharsan can chime in?

Also while running the tests I seem to get a failed one at `test_get_auc` with the error message of ```ValueError: Classification metrics can't handle a mix of multiclass and multiclass-multioutput targets```

nirtiac:
@rfratila  
1. Yep, that's a good point... will do one or the other.
2. No, not a lot of work, will do

nirtiac:
Will close #106

nirtiac:
Closing because all changes are now included in #102

PR #107, opened 2018-12-27
grammar

nirtiac:


nirtiac:
Has been bothering me for a while.

rfratila:
Also if you can remove the old wiki link that would be good for this too please @nirtiac

PR #108, opened 2018-12-28
Include simple example script.

rfratila:
Watered down the multi-input example for a more digestible example to understand Vulcan

### DIFF
--- a/examples/fashion_conv_dense.py
+++ b/examples/fashion_conv_dense.py
@@ -1,26 +1,15 @@
-import sys
-sys.path.append('../')
-from vulcanai2 import models, datasets, plotters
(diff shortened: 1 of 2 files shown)


examples/fashion_conv_dense.py
```diff
@@ -0,0 +1,99 @@
...
+
+import torch
+
+import torchvision.transforms as transforms
+from torch.utils.data import DataLoader, TensorDataset
+# from torchviz import make_dot
+
+
+sys.path.append('../')
+normalize = transforms.Normalize(mean=[x/255.0 for x in [125.3, 123.0, 113.9]],
```
nirtiac: don't think this is actually used... something happened since the first iteration.

rfratila: Good catch! I just pushed a fix to clean up the script a bit more

Issue #95, opened 2018-11-13
network visualization


rfratila:
Using Crayon, or TensorBoardX, or Bokeh, etc

PR #114, opened 2019-01-07
Graph Visualization

priyatharsan:
closes: https://github.com/Aifred-Health/Vulcan/issues/95

inspired from: https://github.com/szagoruyko/pytorchviz/blob/master/torchviz/dot.py


priyatharsan:
I made use of the `baseplotter.py` file which is pretty deserted as of now. But I'm okay with moving to `visualization.py` if needed.

rfratila:
I'm not sure how exactly you made use of baseplotter. The provided code is pretty stand alone and I'm not sure how you would extend it to the other visualizations that we have now. Is there a resource that you used to work on this that you could share? it would make it easier to review. Thanks!

priyatharsan:
The graph visualization is basically inspired from: https://github.com/szagoruyko/pytorchviz/blob/master/torchviz/dot.py (made as simple as possible) apologies that I didn't refer this in description (edited now). and yeah this is pretty stand alone version. But I was thinking if we could customize it in future releases and have it completely integrated with rest of the vulcan's current viz tools. And there is no reason behind why I added within baseplotter, however, I envision that we could improvise in future that encapsulates all the necessary viz within the baseplotter (looking forward to achieve something like this: https://github.com/waleedka/hiddenlayer).

nirtiac:
Closing as a finished alternate is coming.

PR #129, opened 2019-02-13
fix error in identify_sufficient_non_null

nirtiac:
Fixes an error caused during method restructuring. Can wait for if more fixes are found.


vulcanai/datasets/tabulardataset.py
```diff
@@ -385,8 +385,8 @@ def identify_sufficient_non_null(self, threshold):
         num_threshold = (threshold * len(self))
         # thresh is "Require that many non-NA values."
         tmp = self.df.dropna(thresh=num_threshold, axis=1)
-        cols = list(set(self.df.columns).difference(set(tmp.columns)))
-        return cols
+        cols = tmp.columns
+        return list(cols)
```
rfratila: Could you explain the difference in the change?

PR #130, opened 2019-02-13
Metrics fix

nirtiac:
Allows for binary data within run_test. This involved changing the macro calculations for a number of metrics. 

@rfratila while I thought through and googled these to make sure it was logically sound to just take the average in the binary case, I'm not 100%. Could you take a quick look?


vulcanai/models/metrics.py
```diff
@@ -243,7 +243,7 @@ def get_specificity(targets, predictions, average=None):
         specificity = tn / (tn + fp)
 
         # TODO: implement other options
-        if average == "macro":
+        if average == "macro" or average == "binary":
             specificity = np.average(specificity)
```
rfratila: I don't think it's a good idea to average a binary sensitivity or specificity since really it's only referring to the positive class you are trying to predict. @nirtiac @joemehltretter is there a way to standardize which class is the positive one and in the binary case return the corresponding value of the positive class?

nirtiac: Hmmmmm ok yeah. However, I'm reluctant to standardize unless we make it absolutely bulletproof as I see that resulting in uncaught errors with serious implications. We could simply just return None for those cases.

rfratila: yeah I can see that being problematic and harder to use. perhaps there shouldn't be any averaging for binary cases?

nirtiac: As a counterpoint to myself, take a look at what scikit does for recall score (I'm emulating skl_metrics as much as possible in this class). 
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.recall_score.html

rfratila: I do like the pos_label quality if an only if binary is used. + lots of error checking and verbosity in the documentation

nirtiac:
I agree. I wonder how we can implement this into run_test though? We’d have to propagate the parameter back up to something user-specified. Any ideas?

On Feb 13, 2019, at 5:00 PM, Robert Fratila <notifications@github.com<mailto:notifications@github.com>> wrote:


@rfratila commented on this pull request.

________________________________

In vulcanai/models/metrics.py<https://github.com/Aifred-Health/Vulcan/pull/130#discussion_r256607520>:

> @@ -243,7 +243,7 @@ def get_specificity(targets, predictions, average=None):
         specificity = tn / (tn + fp)

         # TODO: implement other options
-        if average == "macro":
+        if average == "macro" or average == "binary":
             specificity = np.average(specificity)


I do like the pos_label quality if an only if binary is used. + lots of error checking and verbosity in the documentation

rfratila:
@nirtiac since it is a very specific function, what if we have a *kwargs for `run_test`? it may not be necessary to have a `pos_label` in the method signature

nirtiac:
yeah I like that @rfratila. Will implement, but it'll take a bit longer than today.  And if it's binary data and they don't provide pos_label it throws an error? I'll make it the same as scikit. 
@joemehltretter you can use it now just fine, but ignore any outputs for the "macro" values.

joemehltretter:
@nirtiac All macro values or just sensitivity and specificity? The main one I've been looking at is macro_auc.

nirtiac:
ummmm... macro_auc is calculated as the average of the auc value for a single class. So if that applies in the binary case then yes, take a look at it?

rfratila:
For the binary case I would just look at the AUC of the positive class @joemehltretter . The Macro arc makes more sense for strictly >2 classes. @nirtiac yeah just keeping it similar to scikit-learn will be familiar with most people. We can add a logger warning just so people know what is being held as the positive label.

nirtiac:
@rfratila I want to think over what I did with auc tomorrow with a fresh pair of eyes

vulcanai/models/basenetwork.py
```diff
@@ -680,13 +680,15 @@ def _validate(self, val_loader):
 
         return validation_loss, validation_accuracy
 
-    def run_test(self, data_loader, plot=False, save_path=None):
+    def run_test(self, data_loader, plot=False, save_path=None, pos_label=1):
```
rfratila: the one thing I'm worried about is what if the user is predicting multiple positive labels in one instance?

nirtiac: I'm not sure I understand.... this is only for the binary case. (which reminds me that I need to add that to the documentation)

rfratila: so the pos_label is only used for the binary case? let's just make sure that is clear in the docs, thanks!

nirtiac: yeah, I'll do that. does everything else look ok? can you take a look at the diff for the auc function?

rfratila: yup ill take a look!

vulcanai/models/metrics.py
```diff
@@ -427,27 +451,25 @@ def get_auc(targets, raw_predictions, num_classes, average=None):
 
         all_class_auc = []
         for i in range(num_classes):
-            if num_classes == 1:
-                fpr, tpr, _ = skl_metrics.roc_curve(targets,
-                                                    raw_predictions,
-                                                    pos_label=1)
-            else:
-                fpr, tpr, _ = skl_metrics.roc_curve(targets,
+            fpr, tpr, _ = skl_metrics.roc_curve(targets,
```
rfratila: Since we technically could be getting two forms of binary inputs (e.g. [[0, 1],[1, 0]] and [1, 0]), this was to handle those 2 cases. @nirtiac would the pos_label generalize to these cases? I would love for us to be able to condense it but I want to be absolutely sure that it's still fine before we do that.

nirtiac: ahhhhh no it doesn't necessarily..... will revert. have been testing with just the one case.

vulcanai/models/basenetwork.py
```diff
@@ -51,7 +51,7 @@ class BaseNetwork(nn.Module):
         input_networks : list of BaseNetwork
             A network object provided as input.
         num_classes : int or None
-            The number of classes to predict.
+            The number of classes to predict. Must be at least 2.
```
rfratila: Would you be able to add more descriptions here on how a user would encode a binary outcome as 2 classes just so there's no confusion?

rfratila:
Before we merge, @joemehltretter would you be able to review this PR?

vulcanai/models/basenetwork.py
```diff
@@ -99,6 +100,8 @@ def __init__(self, name, config, in_dim=None, save_path=None,
         else:
             self.input_networks = input_networks
 
+        if num_classes and num_classes < 2:
```
rfratila: @nirtiac Actually I just realized there may be a big problem with this approach. What if the user wants to predict a single continuous value? Would they just end off on a 1 hidden unit? that may not work since there would be an activation at the end when you woo'dn't want one for continuous value prediction.

nirtiac:
@rfratila 

This includes the changes we discussed. Only thing I didn't add is anything to do with rounding single continuous output values. We wouldn't want to do this for evaluation or for loss, yeah? Just for presentation or output in the final product?

vulcanai/models/basenetwork.py
```diff
@@ -166,52 +170,32 @@ def _add_input_network(self, in_network):
...
+                            nn.NLLLoss and softmax for improved efficiency, \
+                            you cannot set pred_activation when \
+                            criter_spec is set to an instance of this class. \
+                            Set pred_activation to none or change criter_spec."
+                                 )
 
-        """
-        if not isinstance(inputs, list):
-            inputs = [inputs]
+                self._pred_activation = None
```
rfratila: This line of code is unreachable

vulcanai/models/basenetwork.py
```diff
@@ -166,52 +170,32 @@ def _add_input_network(self, in_network):
...
 
-    @abc.abstractmethod
-    def _merge_input_network_outputs(self, inputs):
-        """Abstract method used to define how to handle multi-inputs."""
-        raise NotImplementedError
-
-    def forward(self, inputs, **kwargs):
+    def _set_final_layer_parameters(self, pred_activation, criter_spec):
+        """
+        Sets and checks the parameters used in the final output layer. Final
```
rfratila: Can we clean up this doc so it matches the rest of the doc conventions?

vulcanai/models/metrics.py
```diff
@@ -109,7 +109,7 @@ def extract_class_labels(in_matrix):
         # For one-hot encoded entries
         if in_matrix.shape[1] > 1:
             return np.argmax(in_matrix, axis=1)
-        # For binary entries
+        # For single value entries
         elif in_matrix.shape[1] == 1:
```
rfratila: Should we still keep this? perhaps we don't want to round without the user knowing? this was normally for single valued probabilities but continuous values would fall under this category

vulcanai/models/metrics.py
```diff
@@ -440,14 +464,71 @@ def get_auc(targets, raw_predictions, num_classes, average=None):
...
+        Parameters:
+            targets: numpy.ndarray of integers
+                The target values.
+            raw_predictions: numpy.ndarray of floats
+                The raw predicted values, not converted to classes.
+
+        Returns:
+              results:dict
+        """
+
```
rfratila: can you remove this line to satisfy pylint?

vulcanai/models/metrics.py
```diff
@@ -440,14 +464,71 @@ def get_auc(targets, raw_predictions, num_classes, average=None):
...
+        return skl_metrics.mean_squared_error(targets, raw_predictions)
 
     @staticmethod
-    def run_test(network, data_loader, plot=False, save_path=None):
+    def run_test(network, data_loader, plot=False, save_path=None,
```
rfratila: Can you add some docs to this function?

vulcanai/models/metrics.py
```diff
@@ -440,14 +464,71 @@ def get_auc(targets, raw_predictions, num_classes, average=None):
...
+
+        if num_classes is None or num_classes == 0:
+            raise ValueError('There\'s no classification layer')
+
+        if num_classes == 1:
+            results_dict = Metrics._run_test_single_continuous(network,
+                                                               data_loader)
+        else:
+            results_dict = Metrics._run_test_multi(network, data_loader,
+                                                plot=plot, save_path=save_path,
```
rfratila: these should be aligned with the line above

vulcanai/models/metrics.py
```diff
@@ -440,14 +464,71 @@ def get_auc(targets, raw_predictions, num_classes, average=None):
...
+                                                pos_label=pos_label)
+
+        return results_dict
+
+    @staticmethod
+    def _run_test_single_continuous(network, data_loader):
```
rfratila: Can you add a tiny bit of docs here?

vulcanai/models/metrics.py
nirtiac: Hmm, well I proposed a solution in my latest commit. You can now provide a callable to forward_pass. I don't think I need to expand it to get_score and run_test, as we'll always just score on non-transformed outputs? Or is that not true and there's a use case fro transforming and then scoring in the single continuous case (I know we do for one hot obvs)?

rfratila: ooooh I really like that transform_callable solution. I do think we should allow the score to be dependent on the transformed outputs but then the transform function would need to be done with Pytorch methods so backdrop can work well. This seems like it would lend itself well to the custom loss functionality no @nirtiac ?

nirtiac: @rfratila yes it does. do we want that in this pull request as well?

rfratila: it might be a little outside the scope of this PR and a little more experimental so if you think we could separate this to another PR than I think that would be better no?

nirtiac: I agree, @rfratila. In that case then I believe that this PR is finally ready.

rfratila: Yes. I would like @joemehltretter to do one last check to see if he is still running into eh bug he was before to make sure it was resolved. Then we are good to go! Thanks

nirtiac: @rfratila I have come to the conclusion that we don't need to write any additional code in order to allow what we've been describing above. All the user would need to do is define a custom pred_activation that calls the same transform_callable function as well as a custom loss function (which well accept outputs passed through pred_activation). I don't think we would use this often enough to justify writing further wrappers, nor do I know what that would be. I could work on an example, however.

rfratila: wait so are you removing the transform_callable?

nirtiac: No, that's still in there in what was convert_to_class and is now transform_outputs (other suggestions accepted)

rfratila: Ah ok and you would be able to pass a modified Softmax transform there to get the class outputs?

nirtiac: yes

rfratila: Ok great. Did we want to have that specific case already implemented so it’s easy to use or nah? @joemehltretter I’d like to hear your thoughts as well on this generalized transform method.

joemehltretter: @rfratila @nirtiac So, `convert_to_class` parameter for `forward_pass` is not `transform_outputs`?

nirtiac: It's not transform_outputs. But if you don't provide a transform_callable then it converts to class? Is that too opaque?

joemehltretter: What is an example of `transform_callable`? Just so I can be clear in my understanding

rfratila: @nirtiac I think that might be too much magic to do it automatically. Perhaps there’s something we can do to activate it easily but still manual?

nirtiac: What's "it" in this case @rfratila? @joemehltretter a very simple example is passing np.round. Take a look at the end of predit_single_value_example.py

rfratila: By “it” I’m referring to the class conversion transform by default if nothing is provided

nirtiac: well I can put that back to being a parameter

joemehltretter: The case you gave makes sense, I think it just depends on documentation of it because by itself it may not be easily understood. I do have a `forward_pass` case up and running and ready to test. I set it up last night so can test this tonight.

joemehltretter: The only thing I have to say is we should update the documentation with examples so it's clear for forward_pass so the new parameters are clearly understood. Other than that it's good!

rfratila: @nirtiac would you be able to take care of this ^ thanks :)

nirtiac: dooone @rfratila

rfratila: Would you be able to update the run_test functions? They currently only have:
‘’’
raw_predictions = network.forward_pass(
            data_loader=data_loader,
            convert_to_class=False)
‘’’

nirtiac: ahhhh good catch. they didn't fail cause the basic params are still there, you can provide extras at no consequence.

rfratila: Ah ok I thought we removed convert_to_class

nirtiac: yeah we did, but we have kwargs, so it just ate it up

nirtiac: Ok done, note that I don't propagate any transform_callable to run_test in the multi case cause we need that to be class converted one-hot output values. I did propagate it in the single value case. I did document that, so I think it's good to go. I personally re-tested everything and it all seems reasonable, but I guess it would be good if @joemehltretter used the new and improved run_test.

rfratila: Have you updated the cross validate as well? That calls run test

nirtiac: ack, yes, thanks, done.

rfratila: I did a repo wide search and we have some calls in test_cnn.py, test_dnn.py, test_ensemble.py that would need updating as well.

rfratila: @joemehltretter would you be able to test it now with the new run_test and circle back around to us with any feedback? thanks :)

joemehltretter: @nirtiac @rfratila For `run_test` in `basenetwork` it doesn't match the parameters for `run_test` in metrics, why is that? Because when I call `run_test` after I train my model it doesn't have the transform parameters

nirtiac: ack, you're right. I keep getting overenthusiastic thinking it's done. I'll fix that and do a thorough sweep of all usages everywhere.

nirtiac: Ok, now cross_validate and run_test in basenetwork match in terms of parameters, and my search for all usages of run_test and forward_pass contain the new parameters.

rfratila: Thank you @nirtiac :) @joemehltretter can you confirm that your tests work well with these new updates?

joemehltretter: Just tested, all good.

joemehltretter: Documentation looks good as well

rfratila: Awesome thanks! @nirtiac are there any final cleanups you need to do?

rfratila: I pushed some changes so it looks good to me. If you don't have anything else to do I think it's ready to merge! If you can make sure the last commit is properly tagged so when merged to master (or the master merge commit is tagged) so the pypi build is good, that would be fantastic!

nirtiac: Nope, I'm done. Can you approve so I can merge @rfratila?

PR #135, opened 2019-03-21
Documentation aesthetics

nirtiac:
Some improvements although someone who is ~aesthetic should cut in and tell me what to do. Can be seen here https://vulcanai.readthedocs.io/en/documentation_aesthetics/


nirtiac:
@rfratila updated to use the sticky sidebar, but it doesn't work in Chrome, as they warn. Looks good in firefox thought. I don't think trying to do custom is really worth it at the moment.

rfratila:
Looks good to me @nirtiac!

PR #148, opened 2019-04-23
Selu test fix

joemehltretter:
Fix bug through use of `np.round` in selu weight change test. Since `np.round` was used it sometimes rounded to 0.0 which caused it to fail even if it was technically in the correct range.  Switch to use math.isclose.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ with open('requirements.txt') as f:
 
 setup(
     name='vulcanai',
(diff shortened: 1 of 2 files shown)


nirtiac:
@joemehltretter I just had the misfortune of what I assume is the same error in test_dense_selu_weight_change
Screen Shot 2019-04-29 at 3 23 15 PM

PR #149, opened 2019-04-23
Speedup

nirtiac:
Speedups for everything except TabularDataset, that still needs some fixing. Note, many of the real speedups will have to come from user management of their resources given their data, I'm about to post a wiki article on that.

### DIFF
--- a/vulcanai/datasets/tabulardataset.py
+++ b/vulcanai/datasets/tabulardataset.py
@@ -86,6 +86,9 @@ class TabularDataset(Dataset):
         self.seed_value = int(time.time())
 
         self.set_global_random_seed(self.seed_value)
+
+        self.all_xs = None
+        self.all_y = None
         
         logger.info("You have created a new dataset with %d rows", len(self))
         #logger.info(f"You have created a new dataset with {len(self)} rows")
@@ -130,14 +133,17 @@ class TabularDataset(Dataset):
         # the label column, iloc gets the row, then access values and convert
 
         if self.label_column:
-            xs = self.df.drop(self.label_column,
-                              axis=1).iloc[[2]].values.tolist()[0]
-            xs = torch.tensor(xs, dtype=torch.float)
-            y = self.df[[self.label_column]].iloc[[idx]].values.tolist()[0][0]
+            if self.all_xs is None:
+                self.all_xs = self.df.drop(self.label_column,
+                              axis=1)
+            if self.all_y is None:
+                self.all_y = self.df[[self.label_column]]
+            xs = torch.from_numpy(self.all_xs.iloc[[idx]].values[0]).float()
+            y = self.all_y.iloc[[idx]].values.tolist()[0][0]
             y = torch.tensor(y, dtype=torch.long)
             return xs, y
         else:
-            xs = self.df.iloc[[2]].values.tolist()[0]
+            xs = self.df.iloc[[idx]].values.tolist()[0]
             xs = torch.tensor(xs, dtype=torch.float)
             return xs, None
 
--- a/vulcanai/models/basenetwork.py
+++ b/vulcanai/models/basenetwork.py
@@ -833,6 +833,7 @@ class BaseNetwork(nn.Module):
         else:
             output = torch.cat(inputs, dim=1)
 
+        output = set_tensor_device(output, device=self.device)
         # Return actionable error if input shapes don't match up.
         if output.shape[1:] != self.in_dim:
             raise ValueError(
--- a/vulcanai/models/basenetwork.py
+++ b/vulcanai/models/basenetwork.py
@@ -873,10 +873,11 @@ class BaseNetwork(nn.Module):
         # (e.g. with or without class conversion)
         # so far always a float.
         dtype = torch.float
-        pred_collector = torch.tensor([], dtype=dtype, device=self.device)
+        pred_collector = None
+        #pred_collector = torch.tensor([], dtype=dtype, device=self.device)
         for data, _ in data_loader:
             # Get raw network output
-            raw_outputs = self(data)
+            raw_outputs = self(data).cpu().detach().numpy()
             if self._num_classes:
                 if self._final_transform:
                     predictions = self._final_transform(raw_outputs)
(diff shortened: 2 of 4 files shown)


nirtiac:
Potential serious issue with CUDNN optimization https://stackoverflow.com/questions/48445942/pytorch-training-with-gpu-gives-worse-error-than-training-the-same-thing-with-c

vulcanai/models/basenetwork.py
```diff
@@ -872,32 +873,36 @@ def forward_pass(self, data_loader, transform_outputs=False,
...
-        pred_collector = torch.tensor([], dtype=dtype, device=self.device)
+        pred_collector = None
         for data, _ in data_loader:
+            data = set_tensor_device(data, device=self.device)
             # Get raw network output
-            raw_outputs = self(data)
+            raw_outputs = self(data).cpu().detach().numpy()
             if self._num_classes:
                 if self._final_transform:
                     predictions = self._final_transform(raw_outputs)
```
rfratila: Wouldn't it be better to keep the torch tensor through the final transform?

vulcanai/models/metrics.py
```diff
@@ -62,6 +62,8 @@ def get_score(targets, predictions, metrics='accuracy', average=None,
         functions_without_average_parameter = ["get_accuracy", "get_mse"]
 
         if isinstance(targets, torch.Tensor):
+            # TODO: this is not necessary if targets was created as numpy on
+            # the CPU and then converted using from_numpy
```
rfratila: is there way to do a check for this?

vulcanai/models/metrics.py
```diff
@@ -598,6 +600,7 @@ def _run_test_single_continuous(network, data_loader, transform_outputs,
             results : dict
 
         """
+        # TODO: this may not work with current TabularDataset
         targets = np.array([v[1] for v in data_loader.dataset])
```
rfratila: If you're referring to the matrix stored version within Tabular Dataset, wouldn't that be easier to extract?

vulcanai/datasets/tabulardataset.py
```diff
@@ -130,14 +133,17 @@ def __getitem__(self, idx):
...
-            xs = torch.tensor(xs, dtype=torch.float)
-            y = self.df[[self.label_column]].iloc[[idx]].values.tolist()[0][0]
+            if self.all_xs is None:
+                self.all_xs = self.df.drop(self.label_column,
+                              axis=1)
+            if self.all_y is None:
+                self.all_y = self.df[[self.label_column]]
+            xs = torch.from_numpy(self.all_xs.iloc[[idx]].values[0]).float()
+            y = self.all_y.iloc[[idx]].values.tolist()[0][0]
             y = torch.tensor(y, dtype=torch.long)
```
rfratila: why not also do a `torch.from_numpy` here as well?

vulcanai/models/basenetwork.py
nirtiac: because it could be a numpy function, no?

vulcanai/models/metrics.py
nirtiac: yeah, this'll be resolved as long as we convert it to that. I'll take the comment out.

vulcanai/datasets/tabulardataset.py
nirtiac: because it was giving me all sorts of weird values due to it being a single value (shape was wrong). doesn't matter anyways though because we're going to scrap all of this once I  push the new tabulardataset.

vulcanai/models/metrics.py
nirtiac: hm good point will look into

vulcanai/models/basenetwork.py
rfratila: I guess it could be but this is where we’d apply a soft max for instance and maybe it’s better to keep as torch for optimization?

nirtiac: Only in the case of nn.CrossEntropyLoss?

vulcanai/models/metrics.py
nirtiac: As far as I can tell it appears not @rfratila, although this was interesting reading https://stackoverflow.com/questions/48482787/pytorch-memory-model-torch-from-numpy-vs-torch-tensor

vulcanai/models/basenetwork.py
rfratila: perhaps not only for CrossEntropy but should we restrict users to only Pytorch functions or numpy functions? curious to hear your thoughts on this - I think it really depends on the scope of those final transforms and how complex but still

vulcanai/models/metrics.py
rfratila: perhaps we should modify the condition to make it more useful then? such as checking if on a GPU device?

nirtiac: So I think we want to use .detach regardless cause it adds in requires_grad=False, and then as far as I can tell .cpu is a no-op if it's already on the the CPU, and then we'll have to convert to numpy in any case.

vulcanai/models/basenetwork.py
nirtiac: uhhhhhhhmmmmmm yeah I don't really know. Personally, I think we should choose ease of use over speed, particularly as in this case I don't see a huge GPU advantage? it's a transform after training.

nirtiac: like, with it this way they can give either kind of function. I definitely want to be able to pass something like numpy.round() or whatever.

vulcanai/models/metrics.py
rfratila: sounds good. the condition can then make it support both torch tensors and numpy matrices

vulcanai/models/basenetwork.py
rfratila: You're right - it won't slow down training but then depending on the function passed it may slow it down during inference. To put to ease the speed issue, would you be able to experiment the numpy.round() example and compared with a comparative torch.round() and see what, if any, slow down there is (and potentially for softmax)?

PR #151, opened 2019-04-29
add global random seed

nirtiac:
Works on GPU and CPU, independently..

### DIFF
--- a/vulcanai/__init__.py
+++ b/vulcanai/__init__.py
@@ -24,8 +24,12 @@ logger = logging.getLogger(__name__)
 
 def set_global_seed(seed_value):
     """
-    Sets all the random seeds.
-    :param seed_value: seed value
+    Sets all the random seeds, including for torch, GPU, numpy and python.
+
+    Parameters:
+        seed_value : int
+            The random seed value.
+
     """
     random.seed(seed_value)
     np.random.seed(seed_value)  # cpu vars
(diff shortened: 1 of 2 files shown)


nirtiac:
Also works when combining with the new speedup pull request, provided that num_workers is not set nor is the cudnn backend optimizer.

vulcanai/__init__.py
```diff
@@ -14,3 +20,19 @@
                           defaults={'logfilename': log_output_path})
 
 logger = logging.getLogger(__name__)
+
+
+def random_seeding(seed_value):
```
rfratila: should we call it something like `set_global_seed()`?

nirtiac: sure, that's better!

vulcanai/__init__.py
```diff
@@ -14,3 +20,19 @@
                           defaults={'logfilename': log_output_path})
 
 logger = logging.getLogger(__name__)
+
+
+def set_global_seed(seed_value):
+    """
+    Sets all the random seeds.
+    :param seed_value: seed value
```
rfratila: Can you update this to the more verbose way we do function docs please?

nirtiac: yeah, sorry, bad IDE defaults. done.

rfratila:
It looks good to me so once the tests failures are resolved we’re good to go.

nirtiac:
Just removed tabular dataset's random seed method... will also integrate into the new tabulardataset fix and test conclusively there.

rfratila:
good catch. I don't believe there should be other seed references - but I would double check. I look forward to learn more about your test results!

nirtiac:
note: this conflicts with other current pull requests and needs to be merged last to preserve all tabulardataset changes.

PR #158, opened 2019-05-08
fix wrong parameters

nirtiac:
@joemehltretter I think this may be causing your problem. Sorry :(  The _set_final_layer_parameters function replaces the line that was previously here https://github.com/Aifred-Health/Vulcan/pull/130/files#diff-09d7ff308ec62ab75a95035d00d6c2fcL732 
, however the parameters passed were incorrect, meaning that nothing ever got transformed at the end. 
@rfratila will also add tests so that something like this would be caught in the future.

### DIFF
--- a/vulcanai/models/basenetwork.py
+++ b/vulcanai/models/basenetwork.py
@@ -196,7 +196,7 @@ class BaseNetwork(nn.Module):
                             criter_spec is set to an instance of this class. \
                             Set pred_activation to none or change criter_spec."
                                  )
-            self._final_transform = nn.LogSoftmax(dim=1)
+            self._final_transform = nn.Softmax(dim=1)
 
     @abc.abstractmethod
     def _merge_input_network_outputs(self, inputs):
--- a/vulcanai/models/basenetwork.py
+++ b/vulcanai/models/basenetwork.py
@@ -101,7 +101,8 @@ class BaseNetwork(nn.Module):
         else:
             self.input_networks = input_networks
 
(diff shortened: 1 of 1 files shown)


rfratila:
once the test's are in place we can merge.

joemehltretter:
It looks like it's working for my current models. Definitely producing better results (i.e. Not 100 or 0.0). It's producing some pretty interesting differential probabilities among the drugs. Thanks so much!

joemehltretter:
It did cause one change in the model configuration though, before I was able to able to specify Softmax as the `pred_activation`; however, now it won't let me. It tells me to remove it because of my number of classes.

joemehltretter:
Specifically, this error:

The nn.CrossEntropyLoss class combines nn.NLLLoss and softmax for improved efficiency, you cannot set pred_activation when criter_spec is set to an instance of this class. Set pred_activation to none or change criter_spec.

nirtiac:
@joemehltretter so the error is doing/saying exactly what it's supposed to. I think before you must have been passing your values through softmax twice? If you look at the docs for nn.CrossEntropyLoss then it says  ```This criterion combines nn.LogSoftmax() and nn.NLLLoss() in one single class.``` (side note @rfratila I'm going to have to change the code again... I swear it was nn.Softmax() before.)

joemehltretter:
@nirtiac Hey, so I just tested the most recent commit and its not working. It's producing all negative probabilities some of which are like -100+

PR #164, opened 2019-05-16
Tabular Dataset Improvements

nirtiac:
All new code with the exception of the data storage problem is here, need to test tomorrow. Added new method, cleaned up split and added stratified sampling. 

Not ready for real code review, just concept review :)

### DIFF
--- a/vulcanai/datasets/tabulardataset.py
+++ b/vulcanai/datasets/tabulardataset.py
@@ -554,8 +554,8 @@ class TabularDataset(Dataset):
 
         Returns:
             datasets: Tuple of TabularDatasets
-                Datasets for train, validation, and
-                test splits in that order, if the splits are provided.
+                Datasets for train, test, validation
+                 splits in that order, if the splits are provided.
 
         """
 
@@ -598,6 +598,6 @@ class TabularDataset(Dataset):
         if validation_ratio:
             validation = TabularDataset(val=self.df.loc[validation_indices],
                                         label_column=self.label_column)
-            return train, validation, test
+            return train, test, validation
         else:
             return train, test
--- a/vulcanai/datasets/tabulardataset.py
+++ b/vulcanai/datasets/tabulardataset.py
@@ -581,13 +581,13 @@ class TabularDataset(Dataset):
                 utils.rationed_split(self.df, train_ratio,
                                      test_ratio, validation_ratio)
 
-        train = TabularDataset(train=self.df[train_indices],
+        train = TabularDataset(train=self.df.loc[train_indices],
                                label_column=self.label_column)
-        test = TabularDataset(test=self.df[test_indices],
+        test = TabularDataset(test=self.df.loc[test_indices],
                               label_column=self.label_column)
 
         if validation_ratio:
-            validation = TabularDataset(val=self.df[validation_indices],
+            validation = TabularDataset(val=self.df.loc[validation_indices],
                                         label_column=self.label_column)
             return train, validation, test
         else:
--- a/vulcanai/datasets/tabulardataset.py
+++ b/vulcanai/datasets/tabulardataset.py
@@ -345,7 +345,7 @@ class TabularDataset(Dataset):
         logger.info("Successfully converted %d \
                     columns back from dummy format.", len(dummy_tuples))
 
-    def identify_sufficient_non_null(self, threshold):
+    def identify_null(self, threshold):
         """
         Return columns where there is at least threshold percent of
         null values.
@@ -364,7 +364,7 @@ class TabularDataset(Dataset):
             raise ValueError(
                 "Threshold needs to be a proportion between 0 and 1 \
                 (exclusive)")
-        num_threshold = ((1 - threshold) * len(self))
+        num_threshold = ( (1- threshold) * len(self))
         # thresh is "Require that many non-NA values."
         tmp = self.df.dropna(thresh=num_threshold, axis=1)
         cols = list(set(self.df.columns).difference(set(tmp.columns)))
@@ -373,7 +373,8 @@ class TabularDataset(Dataset):
     def identify_unique(self, threshold):
         """
         Returns columns that do not have have at least threshold number of
-        values
+        values. If a column has 9 values and the threshold is 9, that column
+        will not be returned.
 
         Parameters:
             threshold: The minimum number of values needed.
@@ -385,9 +386,17 @@ class TabularDataset(Dataset):
                 The list of columns having threshold number of values
 
         """
+        obj_types = {col: set(map(type, self.df[col])) for col in
+                     self.df.columns}
+
         column_list = []
         for col in self.df.columns:
-            if len(self.df[col].unique()) <= threshold:
+            if len(obj_types[col]) > 1:
+                logger.warning("Column: {} has mixed datatypes, this may"
+                               "interfere with an accurate identification"
+                               "of mixed values: i.e. you may have 1 and '1'"
+                               .format(col))
+            if len(self.df[col].unique()) < threshold:
                 column_list.append(col)
         return column_list
 
--- a/vulcanai/datasets/tabulardataset.py
+++ b/vulcanai/datasets/tabulardataset.py
@@ -575,14 +575,14 @@ class TabularDataset(Dataset):
 
             grps = self.df.groupby(stratum_column)
             train_index, test_index, val_index = [], [], []
-            for grp in grps:
+            for key, grp in grps:
                 group_train, group_test, group_val = \
                     utils.rationed_split(grp, train_ratio, test_ratio,
                                          validation_ratio)
 
-                train_indices += group_train
-                test_indices += group_test
-                validation_indices += group_val
+                train_indices += list(group_train)
+                test_indices += list(group_test)
+                validation_indices += list(group_val)
 
         else:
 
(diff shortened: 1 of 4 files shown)


nirtiac:
Yes, the tests for those that I reversed are failing :) fixes coming soon.

vulcanai/datasets/tabulardataset.py
```diff
@@ -473,10 +475,49 @@ def identify_low_variance(self, threshold):
                     dct_low_var[col] = col_var
         return dct_low_var
 
+    def convert_all_categorical_binary(self, list_only=False,
```
rfratila: @nirtiac shouldn't this be easily extendable to n>2 categorical types? similar to how the cat codes works in pandas? or is this specifically avoided?

vulcanai/datasets/tabulardataset.py
```diff
@@ -473,10 +475,49 @@ def identify_low_variance(self, threshold):
...
+                self.df = self.df.replace({col: di})
+            except:
+                continue
+            self.df[col].astype(np.float64)
+
     # future improvements could come from
     # https://github.com/pytorch/text/blob/master/torchtext/data/dataset.py
     # noinspection PyUnusedLocal
-    def split(self, split_ratio=0.7, stratified=False, strata_field='label'):
+    def split(self, split_ratio=0.7, stratified=False, strata_column=None):
```
rfratila: should `strata_column` technically be `strati_column` or just `strat_column`?

nirtiac: haha I think it's actually stratum :) will change.

vulcanai/datasets/tabulardataset.py
nirtiac: yeah, that's specifically avoided, we have functions for both categorical and one-hot encoding. One-hot encoding would create two columns, whereas this only creates one. (speaking of, there's a difference between dummy coding and one-hot encoding, where one would create 3 columns for 3 categories and the other 2, as the 3rd column can be 'inferred'. We right now create 3, which is more understandable but somewhat out of line with how we're doing these binary columns) .

rfratila: yeah I see. perhaps there may be some utility iim being able to preserve the one additional column and create dummy codes for n>2?

nirtiac: right now that's what we do for n>2

rfratila: perfect. thanks for the clarification.

nirtiac:
@rfratila this is ready for real review. This is an intermediate push to restore functionality for all methods for now, before any more intensive updates are made to the data storage behind the class.

vulcanai/datasets/tabulardataset.py
```diff
@@ -345,14 +345,15 @@ def reverse_create_one_hot_encoding(self, column_name=None,
         logger.info("Successfully converted %d \
                     columns back from dummy format.", len(dummy_tuples))
 
-    def identify_sufficient_non_null(self, threshold):
+    def identify_null(self, threshold):
```
rfratila: this restructuring and rethinking of this function makes a lot more sense. Great work!

vulcanai/datasets/tabulardataset.py
```diff
@@ -498,40 +548,55 @@ def split(self, split_ratio=0.7, stratified=False, strata_field='label'):
...
-            test_start = train_end
-        test_df = self.df.loc[perm[test_start:]]
 
-        train = TabularDataset(train=train_df, label_column=self.label_column)
-        test = TabularDataset(test=test_df, label_column=self.label_column)
+            train_indices, test_indices, validation_indices = \
+                utils.rationed_split(self.df, train_ratio,
+                                     test_ratio, validation_ratio)
+
+        train = TabularDataset(train=self.df.loc[train_indices],
```
rfratila: @nirtiac would you want to also include the batch parameter in this since the user can't really pass that information here?

vulcanai/datasets/utils.py
```diff
@@ -66,83 +67,23 @@ def check_split_ratio(split_ratio):
...
-#     @random_state.setter
-#     def random_state(self, s):
-#         self._random_state = s
-#
-#     def __call__(self, data):
-#         """Shuffle and return a new list."""
-#         with self.use_internal_state():
-#             return random.sample(data, len(data))
+
+def rationed_split(df, train_ratio, test_ratio, validation_ratio):
```
rfratila: perhaps the way this function is used we should modify the signature to be private?

vulcanai/tests/datasets/test_tabulardataset.py
```diff
@@ -80,15 +80,15 @@ def test_reverse_create_all_one_hot_encodings(self, my_test_dataset):
...
+    def test_identify_null(self, my_test_dataset):
+        num_threshold = 0.2
+        res = my_test_dataset.identify_null(num_threshold)
+        assert {'fedyrs'} == set(res)
 
     def test_identify_unique(self, my_test_dataset):
         res = my_test_dataset.identify_unique(5)
-        assert set(res) == {'id','length','Birthweight','Gestation','motherage','mnocig','mheight','mppwt','fage'}
+        assert set(res) == {'headcirumference', 'smoker', 'fedyrs', 'LowBirthWeight'}
+
```
rfratila: did you want to add a test for the split functionality including the stratified component?

vulcanai/datasets/utils.py
nirtiac: mmmm yes in theory, but not when it's a member of utils? then I think it should be public because utilities implies it's a function anyone can use for whatever. There is however an argument to get rid of the utils file entirely. I'm not the biggest fan, as it seems somewhat arbitrary.

vulcanai/datasets/tabulardataset.py
nirtiac: mmm, that applies only to dataloader, no?

rfratila: true - my mistake

vulcanai/datasets/utils.py
rfratila: ah yes I seem two have missed it's location in `utils.py`. I think it's useful enough to be used independently and thus merits having it as a public function. In that case, would you be able to add some documentation for this function?

vulcanai/tests/datasets/test_tabulardataset.py
```diff
@@ -108,4 +107,31 @@ def test_split_length_correct(self, my_test_dataset):
...
+    def test_split_length_correct_stratified(self, my_test_dataset):
+        res = my_test_dataset.split([0.1, 0.2, 0.7], stratified=True, stratum_column=
+                                             "motherage" )
+        assert len(res) == 3
+        assert isinstance(res[1], TabularDataset)
+        res = my_test_dataset.split([0.1, 0.9], stratified=True, stratum_column=
+                                             "motherage")
+        assert len(res) == 2
+        assert isinstance(res[1], TabularDataset)
+
```
rfratila: you can probably shorten up this gap between the tests for the pylinter

nirtiac:
@joemehltretter would you mind doing a quick review based on your experience using stratified splitting? Just if you could verify that it all works as expected for you.

vulcanai/datasets/tabulardataset.py
```diff
@@ -473,10 +484,49 @@ def identify_low_variance(self, threshold):
                     dct_low_var[col] = col_var
         return dct_low_var
 
+    def convert_all_categorical_binary(self, list_only=False,
+                                       exception_columns = None):
+        """
+        Recodes all columns with only two values as ones and zeros, float
```
rfratila: The one line summary should be separated from the long form description

vulcanai/datasets/tabulardataset.py
```diff
@@ -473,10 +484,49 @@ def identify_low_variance(self, threshold):
                     dct_low_var[col] = col_var
         return dct_low_var
 
+    def convert_all_categorical_binary(self, list_only=False,
+                                       exception_columns = None):
```
rfratila: can you remove the whitespace around the equals?

vulcanai/datasets/tabulardataset.py
```diff
@@ -473,10 +484,49 @@ def identify_low_variance(self, threshold):
...
+        Parameters:
+             list_only: boolean
+                only return a list of columns for which this would
+                apply, do not actually do the transformation
+            exception_columns: list
+                list of column names you do not wish to convert.
+
+        Returns:
+            list or None
+            list if list_only if true, nothing otherwise.
```
rfratila: you should mention it does it inlace if false just for clarity

vulcanai/datasets/utils.py
```diff
@@ -66,83 +67,42 @@ def check_split_ratio(split_ratio):
...
+        test_ratio: int
+            proportion of the dataset that will go to the test split.
+            between 0 and 1
+        validation_ratio: int
+            proportion of the dataset that will go to the val split.
+            between 0 and 1
+
+    Returns:
+        indices: tuple of list of indices.
+    """
```
rfratila: a quick test or assertion should be done that the ratios add to 1

nirtiac: Done before in check_split_ratio when called from split in tabular dataset, but the contract is not explicit. I'd rather modularize and leave that up to the calling function but can be persuaded otherwise. For now, have added docs saying that the check is not performed.

nirtiac: Done before in check_split_ratio when called from split in tabular dataset, but the contract is not explicit. I'd rather modularize and leave that up to the calling function but can be persuaded otherwise. For now, have added docs saying that the check is not performed.

rfratila: Oops I must’ve missed it. Thanks for the clarification!

Issue #131, opened 2019-02-26
pytest missing from requirements


mrkrause:
~/Vulcan/vulcanai/tests/models$ python3 test_cnn.py
Traceback (most recent call last):
  File "test_cnn.py", line 2, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'

PR #169, opened 2019-05-22
add pytest to requirements

nirtiac:
As mentioned in #131


nirtiac:
Yeah, I was thinking about that, we could decide to have a developer requirements file. However, adding this allows that to happen without any additional fiddling.

PR #170, opened 2019-05-24
basic early stopping

nirtiac:
Appears to work, but still need to confirm that the model is being properly copied under all situations/configurations. Creating pull request now for critiques. Admittedly slightly hacky, so feel free to judge.

### DIFF
--- a/vulcanai/models/basenetwork.py
+++ b/vulcanai/models/basenetwork.py
@@ -68,6 +68,8 @@ class BaseNetwork(nn.Module):
             Number of validation iterations of decreasing loss
             (note -not necessarily every epoch!
             before early stopping is applied.
+        early_stopping_metric: string
+            Either "loss" or "accuracy" are implemented.
         criter_spec : dict
             criterion specification with name and all its parameters.
         device : str or torch.device
@@ -85,6 +87,7 @@ class BaseNetwork(nn.Module):
                  optim_spec={'name': 'Adam', 'lr': 0.001},
                  lr_scheduler=None, early_stopping=None,
                  early_stopping_patience=None,
+                 early_stopping_metric = "accuracy",
                  criter_spec=nn.CrossEntropyLoss(),
                  device="cuda:0"):
         """Define, initialize, and build the BaseNetwork."""
@@ -114,6 +117,7 @@ class BaseNetwork(nn.Module):
         self._lr_scheduler = lr_scheduler
         self._early_stopping = early_stopping
         self._early_stopping_patience = early_stopping_patience
+        self._early_stopping_metric = early_stopping_metric
 
         if self._num_classes:
             self.metrics = Metrics()
@@ -380,6 +384,24 @@ class BaseNetwork(nn.Module):
     def early_stopping_patience(self, value):
         self._early_stopping_patience = value
 
+
+
+    @property
+    def early_stopping_metric(self):
+        """
+        Return the early stopping netric.
+
+        Returns:
+            early_stopping_metric : string
+                The early stopping metric
+
+        """
+        return self._early_stopping_metric
+
+    @early_stopping_metric.setter
+    def early_stopping_metric(self, value):
+        self._early_stopping_metric = value
+
     @property
     def criter_spec(self):
         """
@@ -542,14 +564,11 @@ class BaseNetwork(nn.Module):
             self.best_model = None
             self.save_path = ""
 
-        def __call__(self, val_loss, model):
-
-            # TODO: check this - loss is always negative right?
-            score = -val_loss
+        def __call__(self, score, model):
 
             if self.best_score is None:
                 self.best_score = score
-                self.save_checkpoint(val_loss, model)
+                self.save_checkpoint(score, model)
             elif score < self.best_score:
                 self.counter += 1
                 logger.info('EarlyStopping counter: {} out of {}'.format(
@@ -557,9 +576,10 @@ class BaseNetwork(nn.Module):
                 ))
                 if self.counter >= self.patience:
                     self.early_stop = True
+
             else:
                 self.best_score = score
-                self.save_checkpoint(val_loss, model)
+                self.save_checkpoint(score, model)
                 self.counter = 0
 
         def save_checkpoint(self, val_loss, model):
@@ -632,13 +652,19 @@ class BaseNetwork(nn.Module):
                     self.lr_scheduler.step(epoch=epoch)
 
                 if self.early_stopping:
-                    early_stopping(valid_loss, self)
+                    if self.early_stopping_metric == "loss":
+                        early_stopping(-1* valid_loss, self)
+                    elif self.early_stopping_metric == "accuracy":
+                        early_stopping(valid_acc, self)
+                    else:
+                        raise ValueError("Not a valid stopping metric")
+
                     if early_stopping.early_stop:
                         logger.info("Early stopping")
-                        # this should restore everything to the earlier version
-                        # need to confirm
+                        # restores everything to the earlier saved version
                         self.__dict__.update(self.load_model(
                             early_stopping.save_path).__dict__)
+                        # for tqdm
                         iterator.close()
                         break
 
--- a/vulcanai/models/basenetwork.py
+++ b/vulcanai/models/basenetwork.py
@@ -543,18 +543,18 @@ class BaseNetwork(nn.Module):
     # /pytorchtools.py
     class EarlyStopping:
         """Early stops the training if validation loss doesn't improve after
-        a given number of epochs."""
+        a given number of epochs.
+        Args:
+            patience: int default 7
+                How many epochs to wait after last time
+                validation loss improved.
+            verbose: bool default False
+                If True, prints a message for each
+                validation loss improvement.
+        """
 
         def __init__(self, patience=2, verbose=False):
-            """
-            Args:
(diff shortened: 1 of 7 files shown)


vulcanai/models/basenetwork.py
```diff
@@ -494,6 +516,61 @@ def _assert_same_devices(self, comparison_device=None):
                     comparison_device,
                     incompatible_collector))
 
+    # modified from
+    # https://github.com/Bjarten/early-stopping-pytorch/blob/master
+    # /pytorchtools.py
+    class EarlyStopping:
+        """Early stops the training if validation loss doesn't improve after
```
rfratila: would there be a use case to use the validation accuracy as the stopping metric instead of the loss? just seeing if there's a way to generalize this metric unless this is the standard way of identifying early stopping procedures

nirtiac: I haven't a clue what is standard. I can make it a parameter, though, that's probably worthwhile yes?

rfratila: Yeah having it parametrized is ideal

vulcanai/models/basenetwork.py
```diff
@@ -550,6 +651,23 @@ def fit(self, train_loader, val_loader, epochs,
...
+                        early_stopping(-1* valid_loss, self)
+                    elif self.early_stopping_metric == "accuracy":
+                        early_stopping(valid_acc, self)
+                    else:
+                        raise ValueError("Not a valid stopping metric")
+
+                    if early_stopping.early_stop:
+                        logger.info("Early stopping")
+                        # restores everything to the earlier saved version
+                        self.__dict__.update(self.load_model(
```
rfratila: is there some test we can write that would capture this change works as intended? perhaps a very small toy model that deterministically learns the same way every time and we can trigger such event @nirtiac ?

vulcanai/models/basenetwork.py
```diff
@@ -494,6 +538,59 @@ def _assert_same_devices(self, comparison_device=None):
...
+                self.save_checkpoint(score, model)
+                self.counter = 0
+
+        def save_checkpoint(self, val_loss, model):
+            '''Saves model when validation loss decrease.'''
+            if self.verbose:
+                logger.info(
+                    f'Validation loss decreased ({self.val_loss_min:.6f} --> '
+                    f'{val_loss:.6f}).  Saving model ...')
+            self.save_path = model.save_model()
```
rfratila: is there a specific reason that the early stopping checkpoint is saved and not just stored in memory? I guess it's better incase the script suddenly stops but wondering what your thoughts were on this.

nirtiac: yeah that's why I did it, although a downside is slightly increased latency

vulcanai/models/basenetwork.py
nirtiac: yeah but that doesn't really test if we've kept the correct one. but. sure.

rfratila: True - In that case how about we keep track of the best model outside the training process and then take a hash of the model. Then once the training is complete you can do a simple checksum of the model that was self selected and see if it matches exactly the one that the external method found as well? Shouldn’t be too demanding if it’s a small toy model

nirtiac: had thought of that, but https://discuss.pytorch.org/t/why-does-the-hash-value-of-models-change-everytime-i-save-it/24249

nirtiac: so would only work if it's truly the same object. I tested it by comparing all parameters (manually), could do something similar.

rfratila: Ah yes - very good point. It makes sense to reserve the checks for just the parameters

nirtiac: actually, no can do, as we don't have control or access to the model at the time when the best model is found.

rfratila: perhaps if we set up a hook and trigger a pull under some condition we can access it? similar to how the GuidedBackprop works?

nirtiac: I think that's too much to set up just for a test imo

rfratila: Fair enough. Is there some other alternative that you think we can test with this? What if we train two separate but identically initialized models and see if the final versions are the same?

vulcanai/tests/models/test_cnn.py
```diff
@@ -217,3 +217,27 @@ def test_forward_pass_class_not_nan_single_value(self,
...
+            train_loader=ds,
+            val_loader=ds,
+            epochs=5)
+
+        conv3D_net_class_early_stopping.fit(
+            train_loader=ds,
+            val_loader=ds,
+            epochs=5)
+
+        stopping_params = list(conv3D_net_class_early_stopping.parameters())[-1].data
```
rfratila: is setting the torch seed enough or should we use our global seed to ensure they train in exactly the same fashion, otherwise this test will always pass regardless if early stopping works. Same goes for the DNN test

vulcanai/models/basenetwork.py
```diff
@@ -494,6 +538,59 @@ def _assert_same_devices(self, comparison_device=None):
...
+    # https://github.com/Bjarten/early-stopping-pytorch/blob/master
+    # /pytorchtools.py
+    class EarlyStopping:
+        """Early stops the training if validation loss doesn't improve after
+        a given patience."""
+
+        def __init__(self, patience=2, verbose=False):
+            """
+            Args:
+                patience (int): How long to wait after last time validation
```
rfratila: by "how long" imi assuming you are referring to the # of epochs passed? Let's clarify that iim the docs please

vulcanai/models/basenetwork.py
```diff
@@ -566,10 +566,20 @@ def __init__(self, patience=2, verbose=False):
...
+            if not score:
+                logger.info('EarlyStopping counter not incremented, score'
+                            'was not calculated this epoch')
+
+            elif score and not np.isnan(score) and self.best_score is None:
                 self.best_score = score
                 self.save_checkpoint(score, model)
-            elif score < self.best_score:
+
+            # never save a model that has nan as a score, but count
```
rfratila: very clever!

vulcanai/models/basenetwork.py
```diff
@@ -646,7 +656,7 @@ def fit(self, train_loader, val_loader, epochs,
                 train_loss, train_acc = self._train_epoch(train_loader,
                                                           retain_graph)
 
-                valid_loss = valid_acc = np.nan
+                valid_loss = valid_acc = None
```
rfratila: nice - it's good to have this updated since we had such trouble deciding what to leave it as. However, I think there may be issues when it comes to plotting the curves with matplotlib. I would double check that the val loss still lines up with the epochs that it's actually calculated at

nirtiac: yep it destroys the tests. now fixed by converting back to np.nan after the early stopping tests are done... refactor suggestions appreciated, although it's fine as is

rfratila: It's something specific with how matplotlib prints it so when we migrate to a better visualization library we might be able to handle this better so it's not immediately alarming because nan is not a great place holder for score values in the world of DL :P

nirtiac: haha yep agreed

vulcanai/tests/models/test_cnn.py
```diff
@@ -217,3 +217,53 @@ def test_forward_pass_class_not_nan_single_value(self,
...
         assert np.any(~np.isnan(raw_output))
+
+    def test_early_stopping(self, conv3D_net_class_early_stopping,
+                            conv3D_net_class):
+        """ Test that their final params are different: aka
+        that the early stopping did something.
+        Constant seed resetting due to worker init function of dataloader
+        https://github.com/pytorch/pytorch/issues/7068 """
+
+        vulcanai.set_global_seed(42)
```
rfratila: does it really need to be set this many times?

vulcanai/models/basenetwork.py
```diff
@@ -494,6 +538,70 @@ def _assert_same_devices(self, comparison_device=None):
...
 
+    # modified from
+    # https://github.com/Bjarten/early-stopping-pytorch/blob/master
+    # /pytorchtools.py
+    class EarlyStopping:
+        """Early stops the training if validation loss doesn't improve after
+        a given number of epochs."""
+
+        def __init__(self, patience=2, verbose=False):
+            """
```
rfratila: I think these class docs should be one layer above with the class description

vulcanai/tests/models/test_cnn.py
nirtiac: it does when wanting to get determinacy creating several dataloaders, as they're each drawing from random. Open to other suggestions, I don't see one at the moment however.

rfratila: So we get identical data loaders got it. Yeah I’m sure it’s good.

nirtiac:
Cool - I think a full pass-through of everything is needed, not just this commit. We've been letting that drop a bit :P

rfratila:
Yeah I think there’s a pylinter addon or something to the list of checks which would be useful and help with keeping it clean :)

nirtiac:
yeah agreed, will add it to the improvements list.

PR #177, opened 2019-06-19
Very Basic Implementation of Stratified K-Fold Cross Validation

nirtiac:


joemehltretter:
Should the stratified and stat column parameter also be in the `basenetwork` `cross_validate` function?

nirtiac:
Yep, they should indeed, I'll add that in, although it's working now thanks to kwargs.

vulcanai/models/metrics.py
```diff
@@ -892,12 +893,41 @@ def _boot_cv(network, data_loader, k, epochs, retain_graph, valid_interv,
             logger.info(
                 "\n\n***KeyboardInterrupt: Cross validate stopped \
                 prematurely.***\n\n")
-	
+
+    @staticmethod
+    def stratified_split(dataset, k, strata_column):
+
```
rfratila: should we add some assertions about the type of these variable? might as well add a tiny but of docs here

vulcanai/models/metrics.py
```diff
@@ -892,12 +893,41 @@ def _boot_cv(network, data_loader, k, epochs, retain_graph, valid_interv,
...
+            #just let it fail if it doesn't work
+            sr = pd.Series(dataset[0][strata_column].numpy())
+
+        ls = [[] for i in range(k)]
+
+        grps = sr.groupby(by=sr)
+
+        for key, grp in grps:
+            inds = np.array(grp.index)
+            np.random.shuffle(inds)
```
rfratila: just want to double check this is done in-place?

nirtiac: it defs is.... because I also assume it's not and then mess everything up. not this time!

rfratila: amazing! glad to hear it :)

rfratila:
@nirtiac would you be able to add some tests for this function?

vulcanai/models/metrics.py
```diff
@@ -903,11 +904,57 @@ def _boot_cv(network, data_loader, k, epochs, retain_graph, valid_interv,
...
+            A list of dataset objects split into k parts.
+
+        """
+
+        if strata_column == "class_label":
+            sr = pd.Series(dataset.tensors[1].numpy())
+        else:
+            # just let it fail if it doesn't work
+            # does not work for complex multidimensional data
+            sr = pd.Series(dataset.tensors[0][strata_column].numpy())
```
rfratila: sorry, why are you taking the 0-ith vs 1-st index of the tensors? and to clarify we will let it fail if the num of dims is >2?

nirtiac: because the 0th is the data and the 1st is the target. I've added comments to that effect. 
and, yes, we will let it fail.

rfratila: any way to generalize to multiple target tensors?

nirtiac: I think that's out of scope of this function and something that isn't supported in any of our current vulcan implementation, unless I'm mistaken

rfratila: fair enough - you're right

PR #180, opened 2019-06-21
add changes

nirtiac:


vulcanai/plotters/visualization.py
```diff
@@ -215,7 +216,7 @@ def _plot_reduction(x_transform, targets, label_map, title, save_path=None):
             "Display environment needed to plot, "
             "or set save_path=path/to/dir")
     else:
-        plt.show(False)
+        plt.show(block=False)
```
nirtiac: I believe that this will change the functionality: we want to have plots show up while the model is training, although this does depend heavily on a users' environment. Is there another way to get rid of whatever warning this is related to? https://stackoverflow.com/questions/458209/is-there-a-way-to-detach-matplotlib-plots-so-that-the-computation-can-continue/13361748

nirtiac: Repeat for all similar changes below. @rfratila may know more about the intended functionality as I wasn't involved in the writing of this.

vulcanai/tests/plotters/test_visualization.py
```diff
@@ -7,6 +7,11 @@
 from vulcanai.datasets import tabulardataset
 import os
 from copy import deepcopy
+import matplotlib
+import tkinter
+matplotlib.use('TkAgg')
+import warnings
+
```
nirtiac: why are there these new imports?

vulcanai/tests/plotters/test_visualization.py
```diff
@@ -168,6 +172,7 @@ def test_display_pca(self, dnn_class):
 
     def test_display_tsne(self, dnn_class):
         """Test t-SNE displays and saves."""
+        warnings.filterwarnings('ignore', category=PendingDeprecationWarning)
```
nirtiac: can we instead look for something to replace this that's not pending deprecation? we may find ourselves in trouble in the future and may as well figure out how to solve it now.

vulcanai/plotters/visualization.py
rfratila: yes this line is to preserve the plot even once it is no longer training. it is compatible with running python interactively (i.e. in the command line you can run `python -i ...`) to keep the plots up even when the script finishes. More importantly, this is to keep the rest of the script from stalling so it doesn't force you to close the plot in order to continue running the rest of the script.

vulcanai/tests/models/test_metrics.py
```diff
@@ -294,13 +293,11 @@ def test_get_auc(self, metrics):
...
         """Test that run_test returns values as expected."""
+        warnings.filterwarnings("ignore", category=sklearn.exceptions.UndefinedMetricWarning)
         num_items = 300
 
-        test_input = torch.Tensor(np.random.randint(0, 10,
-                                                    size=(num_items,
-                                                          *cnn_class.in_dim)))
-        test_target = torch.LongTensor(np.random.randint(0, 10,
-                                                         size=num_items))
+        test_input = torch.Tensor(np.random.randint(0, 10,size=(num_items,*cnn_class.in_dim)))
```
rfratila: the length of this line may exceed to 79 char limit which is why it was formatted as it was

vulcanai/tests/models/test_metrics.py
```diff
@@ -294,13 +293,11 @@ def test_get_auc(self, metrics):
 
     def test_run_test(self, metrics, cnn_class):
         """Test that run_test returns values as expected."""
+        warnings.filterwarnings("ignore", category=sklearn.exceptions.UndefinedMetricWarning)
```
rfratila: is there another way to make sure the warning goes away without simply ignoring it? this applies to all the filter warnings?

divyachandran-ds:
@nirtiac @rfratila 
I will work on all the above suggestions and get back to you.

nirtiac:
Thanks Divya! :) I’m happy to answer any questions during the process.

On Jun 25, 2019, at 12:58 PM, divyachandran-ds <notifications@github.com<mailto:notifications@github.com>> wrote:


@nirtiac<https://github.com/nirtiac> @rfratila<https://github.com/rfratila>
I will work on all the above suggestions and get back to you.

vulcanai/tests/models/test_metrics.py
nirtiac: this is just to deal with the sklearn deprecated stuff

divyachandran-ds:
It was for the interactive display part, it was already added in some other
plots so added it similarly with similar time duration
I will update the requirements

divyachandran-ds:
Hi Caitrin

I have updated the requirements

Regards
Divya.

vulcanai/tests/plotters/test_visualization.py
nirtiac: My comment above is still relevant ^ I believe.

vulcanai/tests/models/test_metrics.py
```diff
@@ -315,11 +313,8 @@ def test_run_test_binary(self, metrics, cnn_class_binary):
         """Test that run_test returns values as expected."""
         num_items = 300
 
-        test_input = torch.Tensor(np.random.randint(0, 2,
-                                                    size=(num_items,
-                                                          *cnn_class_binary.in_dim)))
-        test_target = torch.LongTensor(np.random.randint(0, 2,
-                                                         size=num_items))
+        test_input = torch.Tensor(np.random.randint(0, 2,size=(num_items,*cnn_class_binary.in_dim)))
```
nirtiac: please make sure that every line is within the limit

PR #187, opened 2019-07-18
Sensitivity analysis

sneha-desai:


vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+        the dataset and setting the feature for all examples to every unique value of that feature.
+
+        Parameters
+        ----------
+            network : BaseNetwork
+                Network descendant of BaseNetwork.
+            data_loader : DataLoader
+                A DataLoader object which contains all the data
+            filename: string
+                The name of the file to save the test results to
```
rfratila: I think you are missing the `features` input documentation

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+
+        if network._num_classes is None or network._num_classes == 0:
+            raise ValueError('There\'s no classification layer')
+
+        dataset = data_loader.dataset
+
+        test_df = None
+        col_headers = None
+
+        for feature in range(len(features)):
```
rfratila: would `idx` (aka index) be a better name than feature if you are iterating the the index locations of the features instead of the features themselves? or something like feature_idx?

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+            feature_val_list = dataset[:][0][:, feature].tolist()
+
+            unique_feature_values = list(np.unique(feature_val_list))
```
rfratila: potentially creating a `set` object of this list may be faster (instead of `np.unique`) to get the unique values? @nirtiac what do you say?

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+
+            if len(unique_feature_values) > cutoff:
+                print("================================")
```
rfratila: Please use the logging `logger` functionality similar to what it looks like in the rest of metrics for consistency and greater logging flexibility

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+                print("Cutting off number of features to be tested for feature "
+                      "{} from {} to {}".format(features[feature],
+                                                len(unique_feature_values),
+                                                cutoff))
+                print("================================")
+                unique_feature_values = unique_feature_values[0::int(
```
rfratila: what is this line doing?

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+                        len(unique_feature_values) / cutoff)]
+            dataset_temp = copy.deepcopy(dataset)
```
rfratila: instead of doing a deep copy of `dataset`. Why not deep copy the data loader dataset? then you store one less instance of dataset in memory because you already treat it as immutable

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+            for feature_value in range(len(unique_feature_values)):
+                dataset_temp[:][0][:, feature] = unique_feature_values[
+                    feature_value]
+                print("Testing value {} for feature {}".format(
```
rfratila: can we update this to use the `logger` functionality?

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+                    unique_feature_values[feature_value], features[feature]))
+
+                targets = dataset[:][1].tolist()
+
+                raw_predictions = network.forward_pass(
+                    data_loader=data_loader)
+
+                predictions = self.transform_outputs(raw_predictions)
+
+                tp, _, fp, _ = Metrics.get_confusion_matrix_values(targets,
```
rfratila: why not use the `get_score` function to calculate whatever score you want?

vulcanai/run.py
```diff
@@ -0,0 +1,47 @@
+from torch.utils.data import DataLoader, Subset
```
rfratila: This looks like you made an example script for the sensitivity analysis, awesome! I think it is better suited to be in the `vulcanai/examples` folder with a more informative name, thanks!

vulcanai/run.py
```diff
@@ -0,0 +1,47 @@
...
+from vulcanai.datasets import TabularDataset
+from vulcanai.models import ConvNet, DenseNet
+import sys
+sys.path.append('../')
+from models.metrics import conduct_sensitivity_analysis
+
+def my_test_dataset():
+    """Create a dataset by importing from the test csv"""
+    fpath = "/Users/snehadesai/Documents/Vulcan/vulcanai/tests/datasets/test_data/birthweight_reduced.csv"
```
rfratila: I would use a relative path depending on the actual package and not your installation of Vulcan

vulcanai/tests/models/conftest.py
```diff
@@ -317,3 +317,23 @@ def multi_input_cnn_data(conv2D_net, conv3D_net, multi_input_dnn_data):
         (TensorDataset(torch.rand(size=[10, *conv3D_net.in_dim])), True, False),
         multi_input_dnn_data
     ])
+
+
+@pytest.fixture(scope="module")
+def dnn_class_single_value_2():
```
rfratila: what does the 2 represent? please add a bit more information in the class documentation where this fixture is being used

vulcanai/tests/models/test_metrics.py
```diff
@@ -330,3 +330,12 @@ def test_run_test_binary(self, metrics, cnn_class_binary):
                             'macro_npv', 'macro_f1', 'macro_auc']
 
         assert all(k in res_dict for k in required_metrics)
+
+    def test_sensitivity_analysis(self, metrics, dnn_class_single_value_2,
+                                  sensitivity_data_loader):
+        metrics.conduct_sensitivity_analysis(dnn_class_single_value_2,
```
rfratila: perhaps you can look at the saved file that this function leaves after running to see if the expected outputs are correct. What we have here isn't necessarily enough. You can use the other test csv for it to be a more robust test

vulcanai/tests/models/conftest.py
rfratila: this may change on how you update the test to use the already embedded test csv file

vulcanai/run.py
```diff
@@ -0,0 +1,47 @@
...
+            'dropout': 0.5,
+        },
+        num_classes=22
+    )
+
+
+
+def run_sensitivity_analysis(dnn_class_single_value_2,
+                              tabular_dataset_train_loader):
+    conduct_sensitivity_analysis(dnn_class_single_value_2,
```
rfratila: A thought: Maybe we should return some JSON object along with saving the csv in case we want to programmatically pass the results through an API? @nirtiac I'd also like to get your thoughts on this

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+        if network._num_classes is None or network._num_classes == 0:
+            raise ValueError('There\'s no classification layer')
+
+        dataset = data_loader.dataset
+
+        test_df = None
+        col_headers = None
+
+        for feature in range(len(features)):
+            feature_val_list = dataset[:][0][:, feature].tolist()
```
nirtiac: so in this case the feature could simply be an index? that's a good solution. did you test it with both?

vulcanai/models/metrics.py
nirtiac: yeah set is faster (https://bbengfort.github.io/observations/2017/05/02/python-unique-benchmark.html)

vulcanai/run.py
nirtiac: yes I think we should. we need to standardize this interface though as all our metric functions return something else. so for now just return a python dictionary and save the file but it may be changed in the future.

vulcanai/run.py
sneha-desai: I actually didn't mean to commit this file. It was just for debugging purposes since I couldn't interactively debug. Should I still add it in vulcanai/examples?

vulcanai/models/metrics.py
sneha-desai: Since unique_feature_values are now sorted, it pulls cutoff # of features evenly spaced from the list

sneha-desai: I made a small mistake, will update in the next pr, it needs a math.ceil() instead of an int()

vulcanai/models/metrics.py
sneha-desai: I don't get what you mean? Instead of this line I do data_loader_temp = copy.deepcopy(data_loader)? How will that save memory?

vulcanai/models/metrics.py
sneha-desai: I'm only storing tp and fp though because those are the only ones that matter. The only way to get this would be through get_confusion_matrix(). I could do get_score(...,metrics='confusion_matrix') instead, but is that necessary?

vulcanai/run.py
rfratila: yes if you can clean it up it's worth it

vulcanai/models/metrics.py
rfratila: you set `dataset = dataloader.dataset` and then deep copy `dataset`. Why not just deep copy from `deepcopy(data loader.dataset)` instead? then you don't need to store the data in the local `dataset` variable

vulcanai/models/metrics.py
rfratila: ok let's add a detailed comment about that. it's not immediately obvious what's going on

vulcanai/tests/models/test_metrics.py
sneha-desai: Hm I'm not really sure how to do this... The network weights I have are random for the test

rfratila: It doesn't matter if the network weights are random, as long as you initialize them exactly the same way each time (hint: setting the seed) then you can have deterministic output of the model because your function will operate across the same values every-time in this scenario no? well then once it completes a test run, take a look at the file it saves, validate that it's been correctly calculated, and then hardcode the assertions to match exactly the values you find in the file. You can try to modify the function afterwards which would output a different value which would then fail your assertions. I hope laying it out like this makes sense

vulcanai/run.py
sneha-desai: How should this dict be arranged? Right now, it returns a df with headers {Feature, Value, Number of examples classified as class 0, ... 1, all the way to n (num classes)}. List of dicts per example with these headers as keys? Or something else?

vulcanai/models/metrics.py
sneha-desai: I haven't tested it with both, but it should work just as well. I think I should actually make the default value of features None, such that if it is None, I just create a new list that goes from 1 to size of num columns of data_loader.dataset

vulcanai/tests/models/test_metrics.py
sneha-desai: Yeah it does, thanks for the explanation, I'll update the assertions to reflect this.

vulcanai/run.py
sneha-desai: Ok

vulcanai/models/metrics.py
sneha-desai: I use dataset in other areas in the code (~ lines 1071, 1051). I could just update these to using dataloader.dataset instead of creating the copy beforehand if that's better?

rfratila: yes I think so - that's what other parts of metrics uses when you look around. @nirtiac just want to get a third opinion from you that this is ok?

vulcanai/run.py
nirtiac: oh hmmmm, a dataframe may indeed be more appropriate then. just specify it in the function documentation.

vulcanai/models/metrics.py
nirtiac: yeah I like that!

vulcanai/models/metrics.py
nirtiac: yeah I think you're right Sneha it's good the way it is.

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+                    col_headers = ["Feature", "Value"]
+                    for i in range(num_classes):
+                        col_headers.append("Number of examples classified as "
+                                           "class " + str(i))
+                    test_df = pd.DataFrame(columns=col_headers)
+
+                row_to_append = [features[feature],
+                                 unique_feature_values[feature_value]]
+
+                for i in range(len(tp)):
```
rfratila: @sneha-desai you can use list comprehension and `zip(tp, fp)` to iterate more cleanly through this list

vulcanai/models/metrics.py
```diff
@@ -1012,3 +1014,85 @@ def cross_validate(network, data_loader, k, epochs,
...
+
+                predictions = self.transform_outputs(raw_predictions)
+
+                tp, _, fp, _ = Metrics.get_confusion_matrix_values(targets,
+                                                                   predictions)
+
+                if test_df is None and col_headers is None:
+                    num_classes = network._num_classes
+                    col_headers = ["Feature", "Value"]
+                    for i in range(num_classes):
```
rfratila: you can probably just use `network._num_classes`. no need to reassign a value you already have if you don't change it in-between

sneha-desai:
-

PR #192, opened 2019-08-07
Early stopping: indentation fixes iteration when rounds are skipped

nirtiac:


vulcanai/models/basenetwork.py
```diff
@@ -665,14 +665,16 @@ def fit(self, train_loader, val_loader, epochs,
                     self.lr_scheduler.step(epoch=epoch)
 
                 if self.early_stopping:
-                    if self.early_stopping_metric == "loss":
-                        if valid_loss:
+
+                    # only call if you've actually measured these.
+                    if valid_loss:
+                        if self.early_stopping_metric == "loss":
                             valid_loss_neg = -1 * valid_loss
```
rfratila: you're negating this because early stoping will look for the largest number consistently?

vulcanai/models/basenetwork.py
```diff
@@ -665,14 +665,16 @@ def fit(self, train_loader, val_loader, epochs,
...
-                        early_stopping(valid_loss_neg, self)
-                    elif self.early_stopping_metric == "accuracy":
-                        early_stopping(valid_acc, self)
-                    else:
-                        raise ValueError("Not a valid stopping metric")
+                            early_stopping(valid_loss_neg, self)
+                        elif self.early_stopping_metric == "accuracy":
+                            early_stopping(valid_acc, self)
+                        else:
+                            raise ValueError("Not a valid stopping metric")
```
rfratila: I would briefly also return in the error the two types of valid metrics to use for this to make the error more actionable

vulcanai/models/basenetwork.py
```diff
@@ -665,14 +665,16 @@ def fit(self, train_loader, val_loader, epochs,
...
+                        if self.early_stopping_metric == "loss":
                             valid_loss_neg = -1 * valid_loss
-                        early_stopping(valid_loss_neg, self)
-                    elif self.early_stopping_metric == "accuracy":
-                        early_stopping(valid_acc, self)
-                    else:
-                        raise ValueError("Not a valid stopping metric")
+                            early_stopping(valid_loss_neg, self)
+                        elif self.early_stopping_metric == "accuracy":
+                            early_stopping(valid_acc, self)
```
rfratila: you only check if `valid_loss` is populated, perhaps you should also check if `valid_acc` also is populated?

vulcanai/models/basenetwork.py
nirtiac: yes

vulcanai/models/basenetwork.py

vulcanai/models/basenetwork.py
nirtiac: Yeah, redundant for now given the current structure but may as well add in case something changes

vulcanai/models/basenetwork.py
rfratila: Can we add a comment about that? We’ll continue to add more functionality to be based off of other metrics in the future and this will have to be kept in mind

nirtiac: ok done.

PR #193, opened 2019-08-07
Cleanup

nirtiac:
Cleans up the main code base for syntax, includes some small fixes as specified. Will need to be rebased after the fix_early_stopping is merged.


vulcanai/datasets/tabulardataset.py
```diff
@@ -516,10 +513,12 @@ def convert_all_categorical_binary(self, list_only=False,
             return binary_cols
 
         for col, index in binary_cols:
+            if col in exception_columns:
+                continue
             di = {index[0]: 1.0, index[1]: 0.0}
             try:
                 self.df = self.df.replace({col: di})
-            except:
+            except (AssertionError, TypeError, ValueError):
```
rfratila: perhaps this is too silent of a way to handle these errors?

vulcanai/models/basenetwork.py
```diff
@@ -2,8 +2,6 @@
 """Defines the basenetwork class."""
 # Core imports
 import abc
-import torch
```
rfratila: you're removing these because they are imported on __init__ ?

vulcanai/models/basenetwork.py
```diff
@@ -877,53 +878,43 @@ def run_test(self, data_loader, plot=False, save_path=None, pos_label=1,
...
-	difference in predicting various values for a given feature when predicting outcome. 
-
-	Parameters:
-	    network : BaseNetwork
-                Network descendant of BaseNetwork.
+                            p_output_path=None, **kwargs):
+        """Performs bootfold - estimation to identify whether training model
+        provides statistically significant
+        difference in predicting various values for a given feature when
+        predicting outcome.
```
rfratila: I think you may need an empty line between the description and the parameters?

vulcanai/models/basenetwork.py
```diff
@@ -877,53 +878,43 @@ def run_test(self, data_loader, plot=False, save_path=None, pos_label=1,
...
             epochs : int
                 The number of epochs to train the network per fold.
-	    index_to_iter : string
-		Index of feature within data_loader who's values will be iterated to assess difference
-	    ls_feat_vals : list
-		List of values for feature provided in feat_to_iter
+            index_to_iter : string
+                Index of feature within data_loader who's values will be
+                iterated.
+            to assess difference
```
rfratila: this line needs to be indented?

vulcanai/models/basenetwork.py
```diff
@@ -877,53 +878,43 @@ def run_test(self, data_loader, plot=False, save_path=None, pos_label=1,
     def bootfold_p_estimate(self, data_loader, n_samples, k, epochs, 
                             index_to_iter, ls_feat_vals, retain_graph=None,
                             valid_interv=4, plot=False, save_path=None,
-                            transform_outputs=False, transform_callable=None, p_output_path = None,
```
rfratila: is the transform_callable parameter deprecated?

vulcanai/models/layers.py
```diff
@@ -120,9 +122,7 @@ def __init__(self, in_features, out_features,
         # Main layer
         self._kernel = nn.Linear(
                             in_features=self.in_features,
-                            out_features=self.out_features,
-                            bias=True
```
rfratila: you're removing the `bias=True` because that's the default value right? so it's redundant? just want to make sure

vulcanai/models/layers.py
```diff
@@ -301,6 +301,7 @@ def __init__(self, in_channels, out_channels, bias=False):
 
     def forward(self, input):
```
rfratila: yeah can make sure that this class is hidden so it doesn't automatically get used because it's not close to being ready and it's a bit of a vestigial organ for this file? Thank you

vulcanai/models/metrics.py
```diff
@@ -45,11 +45,12 @@ def get_score(targets, predictions, metrics='accuracy', average=None,
...
             average: string,
-                [None, ‘binary’ (default), ‘micro’, ‘macro’, ‘samples’, ‘weighted’]
+                [None, ‘binary’ (default), ‘micro’, ‘macro’, ‘samples’,
+                ‘weighted’]
                 This parameter is required for multiclass/multilabel targets.
                 If None, the scores for each class are returned.
-                Otherwise, this determines the type of averaging performed on the
-                data.See scikit learn
+                Otherwise, this determines the type of averaging performed on
+                the data.See scikit learn
```
rfratila: should we provide a link or nah?

vulcanai/models/metrics.py
```diff
@@ -615,14 +620,16 @@ def _run_test_multi(network, data_loader, plot=False, save_path=None,
...
+        num_classes = network.num_classes
 
         if num_classes > 2:
             average = "macro"
         elif num_classes == 2:
             logger.warning("Will report scores only for pos_label, which is \
                            set to {}".format(pos_label))
             average = "binary"
+        else:
+            raise ValueError("Incorrect number of classes for run_test_multi")
```
rfratila: I think you should print what was received and what should be expected to make this error more actionable - or at least use that info to further debug

vulcanai/models/utils.py
```diff
@@ -1,65 +1,79 @@
...
+        index_to_iter : string
+            feature to iterate through adjusting values
+        ls_feat_vals : list
+            values to iterate through for feature in index_to_iter
+
     Returns:
-	dct_scores : dictionary
-	    dictionary of scores
-    """
+    dct_scores : dictionary
```
rfratila: this should be indented over

vulcanai/plotters/__init__.py
```diff
@@ -12,6 +8,8 @@
     display_record,
     display_receptive_fields
 )
+import matplotlib
+matplotlib.use('agg')
```
rfratila: we want to make sure the `plt` import happens after the matplotlib backend get's set otherwise it won't change it after `plt` has been imported.

vulcanai/models/basenetwork.py
nirtiac: yes, and not direclty used in this class.

vulcanai/models/basenetwork.py
nirtiac: no yet, but it's not used in this function as Joe implemented it, so I wanted to take it out to avoid thoughts that it was used. I want to get rid of it anyways, in a future pull request.

vulcanai/models/layers.py
nirtiac: yes.

vulcanai/models/metrics.py
nirtiac: I provided further explanation - it's all documented on the individual metric functions so there's not one link.

vulcanai/models/metrics.py
nirtiac: this would never really happen unless someone overwrote the code. but I added more detail as well.

vulcanai/plotters/__init__.py
nirtiac: ok. just reverted .

vulcanai/models/layers.py
nirtiac: done

PR #195, opened 2019-08-15
Refactor

nirtiac:
- cleans up some of the functionality related to training - e.g. we are now by default in self.eval()
- gets rid of the TabularDataset class and replaces it with a collection of functions.


vulcanai/models/basenetwork.py
```diff
@@ -196,17 +194,13 @@ def _set_final_layer_parameters(self, pred_activation, criter_spec):
...
-                raise ValueError("The nn.CrossEntropyLoss class combines  \
+            raise ValueError("The nn.CrossEntropyLoss class combines  \
                             nn.NLLLoss and softmax for improved efficiency, \
-                            you cannot set pred_activation when \
-                            criter_spec is set to an instance of this class. \
-                            Set pred_activation to none or change criter_spec."
-                                 )
-            self._final_transform = nn.Softmax(dim=1)
+                            but is not compatible with our current \
+                            implementation of vulcan"
```
rfratila: I mean... it is compatible except that it becomes more difficult to build ONNX-exportable Vulcan models? let's try to find a better way to word this error or is there something else we can do to make it more inclusive for `CrossEntropyLoss`?

vulcanai/models/basenetwork.py
```diff
@@ -784,6 +775,9 @@ def _train_epoch(self, train_loader, retain_graph):
         train_accuracy = train_accuracy_accumulator * \
             train_loader.batch_size / len(train_loader.dataset)
 
+        # returns the network to evaluation state
+        self.eval()
```
rfratila: shouldn't this be at the end of the entire training loop instead of between epochs?

vulcanai/tests/datasets/test_tabulardataset.py
```diff
@@ -1,144 +1,144 @@
-# coding=utf-8
-""" Defines test cases for tabular dataset """
-import pytest
```
rfratila: woah that's a lot of deletions. Are you working to adapt these tests for the new a-la-carte function structure?

vulcanai/datasets/tabular_data_utils.py
```diff
@@ -0,0 +1,394 @@
...
+from itertools import groupby
+from sklearn import preprocessing
+
+logger = logging.getLogger(__name__)
+
+# TODO: split function
+
+
+# TODO: use kwargs
+def create_label_encoding(df, column_name, ordered_values):
```
rfratila: is there any functionality that we are losing with these split data util functions as compared to the class-based implementation?

vulcanai/datasets/tabulardataset.py
```diff
@@ -1,604 +0,0 @@
...
-    def __len__(self):
-        """
-        The total number of samples.
-
-        Returns:
-            The total number of samples: int
-        """
-        return len(self.df.index)
-
-    def __getitem__(self, idx):
```
rfratila: What are we gonna use to replace this? how will we adapt this refactor to place nicely with the data loader of tabular data? I guess we would need some sort of `convert_tabular_to_dataset` function?

vulcanai/datasets/tabular_data_utils.py
rfratila: also there are a few functions that got removed in this transition around saving dataframes? such as listing features, listing columns, etc

vulcanai/models/basenetwork.py
nirtiac: No it's no longer compatible as it stands because I removed the softmax final_transform parameter. This is something we could add back in, up for debate. I took it out because there's limited benefit for (our) applications and I didn't want to have future oopsies.

vulcanai/models/basenetwork.py
nirtiac: well, there's validation within the training  loop and I wanted .eval() to be the default. We don't need .train() anywhere except here do we?

vulcanai/tests/datasets/test_tabulardataset.py
nirtiac: haha yes, I wrote that in the commit message but should have said it here. I'll add them back once we've decided on the final structure... would rather not spend time writing tests I don't have to

vulcanai/datasets/tabular_data_utils.py
nirtiac: No, not if the user users another split function on their dataframe or on their tabular data. I'd like to move away from us performing splits in any way but very deliberately and manually - generate a test/train/validation set BEFORE anything else happens in the analysis.

nirtiac: yes, I got rid of all the features that were very simple wrappers around simple pandas functions

nirtiac:
This refactor gets rid of tabular dataset and replaces it with a set of utility functions. @joemehltretter what do you think of this? I did this because it was too difficult to go between dataframes and tensors, the current implementation was slow, we weren't using it as intended, and I want users to do all their preprocessing very very separately from model creation. By linking dataframes and models we create a dangerous situation for reproducibility. This forced separation allows us to do things like write intermediate data files, allowing for clearer reproducibility.

vulcanai/models/basenetwork.py
rfratila: oh I see what you mean. let's add that little tidbit of insight to the comment to make it clear why it's here and not elsewhere. thanks!

vulcanai/tests/datasets/test_tabulardataset.py
rfratila: got it - sorry I missed that

vulcanai/datasets/tabular_data_utils.py
rfratila: ok yeah that makes sense. Trimming the unnecessary duplicated fat

nirtiac:
Now unrefactored with regards to the final_transform variable. Was short-sighted: we want to be able to use nn.LogSoftMax for a variety of good reasons, and so will have to transform results using exp or softmax anyways...

rfratila:
Sounds good about the unrefactor. Want to hear back from @joemehltretter about the tabular dataset changes

vulcanai/datasets/tabulardataset.py
nirtiac: yeah, I'll write that

vulcanai/models/basenetwork.py

vulcanai/datasets/tabulardataset.py
nirtiac: done in most recent commit

vulcanai/datasets/tabular_data_utils.py
```diff
@@ -9,12 +9,42 @@
...
+    if not target_vars:
+        return TensorDataset(torch.Tensor(np.array(df)))
+
+    if target_vars and not target_vars.isinstance(list):
+        target_vars = [target_vars]
+
+    data = torch.Tensor(np.array(df.drop(target_vars, axis=1)))
+    target = torch.LongTensor(np.array(df[target_vars]))
+
+    dataset = TensorDataset(data, target)
```
rfratila: did you want to return multiple datasets for the data and all the targets? or do you think it's fine to send it as one?

rfratila: Also what if the target isn't a class but rather a float value? perhaps we should infer the type from the df for the targets?

vulcanai/datasets/tabular_data_utils.py
```diff
@@ -9,12 +9,42 @@
...
+
+    return dataset
```
rfratila: Any tests that you think can accurately capture the function of this method? and related, any error handling we should do if the target_vars doesn't exist in the dataframe (e.g. from a typo)?

vulcanai/datasets/tabular_data_utils.py
nirtiac: that's how we need it to be for training and inference in vulcan

good point about the float values, I'll fix that.

vulcanai/datasets/tabular_data_utils.py
nirtiac: yeah, I'll refactor the tests now.

In general I want to allow errors to be thrown from the specific libraries we're using. so for that we'll get the pandas error and that's ok because it's descriptive enough. The user should receive and deal with that error.

rfratila: I agree with letting the libraries return their own errors, as we will return our own custom ones if pertaining to Vulcan-specific functionality.

vulcanai/datasets/tabular_data_utils.py
```diff
@@ -86,6 +91,22 @@ def create_label_encoding(df, column_name, ordered_values):
...
 
 
+def create_one_hot_encoding(df, column_name, prefix_sep="@"):
+    """
+    Create one-hot encoding for the given column.
+    Parameters:
+        column_name: String
+            The name of the column you want to one-hot encode
+        prefix_sep: String default("@")
+            The prefix used when creating a one-hot encoding
```
rfratila: it may be redundant, but can we add a small "Returns" section in the docs to indicate that the returned df will have a new column that will represent that one hot encoding? or does it stay the same? it would change how the df should be used after passing it through here

vulcanai/tests/datasets/test_tabulardataset.py
```diff
@@ -1,144 +1,186 @@
...
+
+    @pytest.fixture
+    def my_test_dataset(self):
+        """Create a dataset by importing from the test csv"""
+        fpath = str(os.path.dirname(__file__)) + \
+            "/test_data/birthweight_reduced.csv"
+        # Nan just an artifact of this dataset.
+        return pd.read_csv(fpath, na_values='Nan')
+
+    def test_conver_to_tensor_dataset(self, my_test_dataset):
```
rfratila: I think you misspelled `convert`

vulcanai/tests/datasets/test_tabulardataset.py
```diff
@@ -1,144 +1,186 @@
...
+        res = tabular_data_utils.create_one_hot_encoding(my_test_dataset,
+                                                   "LowBirthWeight")
+        assert "LowBirthWeight@Low" in list(res.columns)
+
+    def test_reverse_create_all_one_hot_encodings(self, my_test_dataset):
+        tabular_data_utils.create_one_hot_encoding(my_test_dataset,
+                                                   "LowBirthWeight")
+        res = tabular_data_utils.reverse_create_one_hot_encoding(my_test_dataset,
+                                                           prefix_sep="@")
+        assert "LowBirthWeight@Low" not in list(res.columns)
```
rfratila: you can cast it to a `set` for _slightly_ faster lookup

Issue #202, opened 2023-05-12
'DenseNet' object has no attribute '_num_classes'


zeelsheladiya:
This Error accusing whenever I was trying to do a sentimental analysis from the examples. 

<img width="1440" alt="image" src="https://github.com/Aifred-Health/VulcanAI/assets/46935793/b087acc2-3e5d-419a-8857-3a705af390ef">

I try to go through the functions. it uses _num_classes instead of num_classes

PR #203, opened 2023-05-13
Fixed issue #202

zeelsheladiya:
To resolve this issue, I created a private variable in the DNN class. After doing that, I encountered another error in the `conduct_sensitivity_analysis` function. The error was that the `test_df` dataframe object could not be appended. To fix this, I appended the generated dataframe using the `pd.concat` method, and it worked. I was able to successfully run the test.

dnn.py
image

metrics.py
image

Successful test
image

test CSV file
image


zeelsheladiya:
any input?

